PR #13, opened 2019-05-09
Make Acorn work with Mix's manifest

Log1x:
This adds support for Mix's manifest which requires the path to be prepended with `/`.

When fetching the local path, I made use of `Str::before()` to strip `?id=X` thus making it functional with Mix's built in `.version()` as well as plugins such as [laravel-mix-versionhash](https://github.com/ctf0/laravel-mix-versionhash) which handle hashing via the filename instead of a query string.

### DIFF
--- a/src/Acorn/Application.php
+++ b/src/Acorn/Application.php
@@ -130,6 +130,20 @@ class Application extends Container implements ApplicationContract
     }
 
     /**
+     * Register all of the configured providers.
+     *
+     * @return void
+     */
+    public function registerConfiguredProviders()
+    {
+        collect($this['config']['app']['providers'])->each(
+            function (string $provider) {
+                $this->register($provider);
+            }
+        );
+    }
+
+    /**
      * Register the core class aliases in the container.
      *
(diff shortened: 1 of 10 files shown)


Log1x:
Hm. This ends up being problematic as the slash is stripped in scenarios where the asset you're targeting isn't inside of the Manifest.. 

To be honest, I'm not entirely sure how you want to handle this in the long-run @QWp6t – I know the todo's mention you reworking the Manifest/Asset classes a bit.

Log1x:
Alright.

- Fixed missing prepended slash on files that do not exist in the manifest
- Fixed failing tests
- Added tests for manifest path and URI with query string
- Changed main.js/css to app.js/css in tests to reflect Sage 10
- Added missing trailing space to `!` in a couple conditionals that were missing it

Can't say I'm 100% happy with the `Asset` class, but it's more or less about as clean as I can see it getting at the moment while keeping the ability to support filename revving as well as query strings.

Still cleaner than Laravel's official `mix()` function imo.

Issue #14, opened 2019-05-11
ServiceProviders never initialized.


Log1x:
* [x] I've read the [guidelines for Contributing to Roots Projects](https://github.com/roots/.github/blob/master/CONTRIBUTING.md)
* [x] This request isn't a duplicate of an existing issue
* [x] I've read the [docs](https://roots.io/documentation/) and followed them (if applicable)
* [x] This is not a personal support request that should be posted on the [Roots Discourse](https://discourse.roots.io/) community

## Description

Currently, Acorn does not initialize ServiceProviders defined in `app.providers`

## Steps to reproduce

[...]

## Versions

[...]

## Additional information

[...]

PR #15, opened 2019-05-12
Load config and register providers

kellymears:
Added two private methods to `SageServiceProvider` in order to load the Sage app configuration and register the providers from the `providers` array.

Addresses #14 .

### DIFF
--- a/src/Acorn/Application.php
+++ b/src/Acorn/Application.php
@@ -130,6 +130,20 @@ class Application extends Container implements ApplicationContract
     }
 
     /**
+     * Register all of the configured providers.
+     *
+     * @return void
+     */
+    public function registerConfiguredProviders()
+    {
+        collect($this['config']['app']['providers'])->each(
+            function (string $provider) {
+                $this->register($provider);
+            }
+        );
+    }
+
+    /**
      * Register the core class aliases in the container.
      *
      * @return void
--- a/src/Acorn/Bootstrap/RegisterProviders.php
+++ b/src/Acorn/Bootstrap/RegisterProviders.php
@@ -1,0 +1,19 @@
+<?php
+
+namespace Roots\Acorn\Bootstrap;
+
+use Illuminate\Contracts\Foundation\Application;
+
+class RegisterProviders
+{
+    /**
+     * Bootstrap the given application.
+     *
+     * @param  \Roots\Acorn\Application  $app
+     * @return void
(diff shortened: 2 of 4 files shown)


QWp6t:
ServiceProvider already has a [`mergeConfigFrom()`](https://github.com/laravel/framework/blob/master/src/Illuminate/Support/ServiceProvider.php#L52-L64) method that can be used (sidenote: there's also [`loadViewsFrom()`](https://github.com/laravel/framework/blob/master/src/Illuminate/Support/ServiceProvider.php#L79-L97)). I want there to be a single authority where a user can override any config files, and it needs to be configurable. Right now that's supposed to be \<basepath\>/config, and when Sage is found, the basepath is set to the theme folder. In short, we should use $this->mergeConfigFrom() as needed, but in this case, it's actually not needed since the config file is already getting loaded.

As for app providers not being loaded, that was a bug that I accidentally introduced a couple months ago when I pushed a bunch of updates. It should probably be handled either in a separate [Bootstrap](https://github.com/roots/acorn/tree/master/src/Acorn/Bootstrap) or within Application.php itself ([or](https://github.com/laravel/framework/blob/master/src/Illuminate/Foundation/Application.php#L559-L575) [both](https://github.com/laravel/framework/blob/master/src/Illuminate/Foundation/Bootstrap/RegisterProviders.php)). Feel free to refactor the PR if you want and I'll be happy to merge it. (note: [this is how I used to handle it](https://github.com/roots/acorn/blob/c505e612ae05b1773a1024d115b9e4ecde20cd6c/src/bootstrap.php#L31-L34))

kellymears:
It's straight up a privilege to have a chance to contribute to this awesome thing you've built.

As requested (let me know if I've misinterpreted anything): 

Registration of configurable service providers is removed from `Acorn\Sage\SageServiceProvider` and added to `Acorn\Application`.

Adds a new bootstrap class to call this method: `Acorn\Bootstrap\RegisterProviders`. This bootstrap class has been added to the `$bootstrap` array in the `bootloader()` function of `src/helpers.php`.

With this change configurable service providers are successfully registered and booted based on the `providers` inventory in `config/app.php`.

QWp6t:
Awesome dude! Thanks a bunch for helping with this. I haven't had a chance to revisit the Acorn stuff for a while, but I plan on getting to it sometime next week.

PR #20, opened 2019-06-13
Package Discovery Continued

Log1x:
- Add ComposerScripts class to handle Composer events for package discovery
- Move console commands to their own folder
- Add package:clear command
- Add optimize command
- Add optimize:clear command
- Remove unused use statement in Bindings

Related: https://github.com/roots/sage/pull/2201


QWp6t:
At some point we should look into refactoring ComposerScripts to first attempt to use wp-cli and only fallback to current implementation on failure. That would be cleaner since it wouldn't make any assumptions about Acorn basepath.

PR #21, opened 2019-06-18
[WIP] Check that .blade.php cannot be accessed remotely

QWp6t:
Should this be cached, or should it be a manual check via wp-cli? Should this throw an error or show a more polite notification/alert? What sort of options/filters should be implemented?

Should we instead put this on the ViewServiceProvider and iterate over all view paths within the document root and select an actual .blade.php from each path to ensure all paths are protected?

Not sure how crazy we want to get with this. I'm personally of the mindset that this is outside the scope of Sage and Acorn, and I'd be in favor of completely removing it. But if we do want it, my preference would be to implement it as a wp-cli command.

### DIFF
--- a/src/Acorn/Sage/SageServiceProvider.php
+++ b/src/Acorn/Sage/SageServiceProvider.php
@@ -24,4 +24,35 @@ class SageServiceProvider extends ServiceProvider
             $this->app['sage']->attach();
         }
     }
+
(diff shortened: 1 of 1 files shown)


QWp6t:
I think we're going to create some sort of plugin that adds a [site health check](https://make.wordpress.org/core/2019/04/25/site-health-check-in-5-2/) instead.

PR #22, opened 2019-06-23
Change to CircleCI

Log1x:


QWp6t:
We can have checkout as its own job that persists to the workspace and acts as a requirement for all other jobs. That will make it so that `git checkout` is only run once. Linting and testing can each be their own jobs so that they're run in parallel. These are incredibly minor changes that we can deal with later; I'm just throwing that out there mostly for future reference when one of us gets bored enough to deal with micro-optimizations. 🙊

PR #26, opened 2019-07-13
Add validator support

aplr:
This PR adds support for Laravel's built-in validation component, which did not work before. Since I need the validator in a project of my own, I've fixed that and want to share this.


QWp6t:
Whoops. 🙈

Acorn was initially based more on Lumen than Laravel, and Lumen uses [`Application::getLanguagePath()`](https://github.com/laravel/lumen-framework/blob/master/src/Application.php#L555).

Thanks for the PR!

PR #28, opened 2019-07-15
Use Illuminate\Console for CLI

Log1x:
## Changes

- Move to `Illuminate\Console` with a wrapper for `WP-CLI`
  - Commands found in `Console/Commands` within an app are auto-loaded.
  - `wp acorn make:command <name>`
- Add an `Exception` handler
- Add logging
- Add optimizations to `composer.json`
- Add `whoops` as a dev-dependency (will do more with it soon)
- Various clean up


QWp6t:
i didn't test anything. YOLO

Issue #29, opened 2019-07-16
make:composer command should add new composer to config


knowler:
* [x] I've read the [guidelines for Contributing to Roots Projects](https://github.com/roots/.github/blob/master/CONTRIBUTING.md)
* [x] This request isn't a duplicate of an existing request
* [x] This is not a personal support request that should be posted on the [Roots Discourse](https://discourse.roots.io/) community

## Summary

`wp acorn make:composer` should add new composer to composers in `config/view.php`.

Log1x:
As far as I know, there's no convenient way to write to config files in Laravel's core at the moment. Third party packages do exist for this, however.

I'm kind of thinking we just autoload `Composers` and `Providers` that are in the `app/` directory. We are already autoloading providers from vendors during `composer dump-autoload` as well as autoloading `Commands`– so we might as well + it might be less confusing for people in the long-run.

QWp6t:
I'm fine with that as long as we're not assuming the `App` namespace. There should be a `getNamespace()` helper function ... somewhere. I believe it's on the application container. We use it in a couple commands already.

Log1x:
This is no longer necessary as of #32.

PR #32, opened 2019-07-22
Small clean up for Sage

Log1x:
## Change log

- Use `app.globals` for enabling global helpers instead of a filter    
- Set default timezone using `app.timezone`
- Set default mb encoding to `UTF-8`
- Add docblocks to Sage class
- Clean up Sage class, move to anonymous functions
- Add default comments template (`partials/comments.blade.php`)
- Move Search Form filter from Sage to Acorn
- Move Body Class filter from Sage to Acorn
- **Add Composer autoloading** (#29)

Sage has already been updated to include `app.timezone` and `app.globals`. If you're down with moving the filters over, I'll remove them from Sage as well as remove the specific defining `/partials/comments.blade.php` in partials such as `content-single.blade.php`.

### DIFF
--- a/src/Acorn/Sage/Sage.php
+++ b/src/Acorn/Sage/Sage.php
@@ -71,7 +71,7 @@ class Sage
      */
     public function attach()
     {
-        $this->attachPostVariable();
+        $this->attachPostFilter();
         $this->attachCommentsTemplateFilter();
         $this->attachSearchFormFilter();
         $this->attachBodyClassFilter();
--- a/src/Acorn/Sage/Sage.php
+++ b/src/Acorn/Sage/Sage.php
@@ -61,7 +61,7 @@ class Sage
         $this->files = $files;
         $this->fileFinder = $fileFinder;
         $this->sageFinder = $sageFinder;
-        $this->views = $view;
+        $this->view = $view;
     }
 
     /**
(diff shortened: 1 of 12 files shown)


QWp6t:
I don't want to use anonymous functions because that prohibits theme authors from removing those filters.

src/Acorn/Sage/Sage.php
```diff
@@ -13,53 +13,153 @@
...
     public function attachCommentsTemplateResolver($priority = 100)
     {
-        add_filter('comments_template', [$this, 'commentsTemplateResolver'], $priority);
+        add_filter('comments_template', function ($name) {
+            if (Str::startsWith($file, [STYLESHEETPATH, TEMPLATEPATH])) {
+                $file = ltrim(str_replace([STYLESHEETPATH, TEMPLATEPATH], '', $file), '\\/');
+            }
+
+            if ($file == '/comments.php') {
+                $file = '/partials/comments.blade.php';
```
QWp6t: This should be `'partials/comments.blade.php'` to be consistent with `ltrim()` above and because it's a relative path.

Log1x:
All of the filters have been separated into their own functions as well as cleaned up, documented, etc.

I have also added View Composer autoloading which subsequently solves #29 and will remove a step when it comes to working with Sage, etc.

QWp6t:
I removed the following changes:
* Set default timezone using app.timezone: We can discuss in separate issue/pr if you want, but I think Bedrock would be the more appropriate place for this.
* Set default mb encoding to UTF-8: [WordPress already handles this](https://github.com/WordPress/WordPress/blob/master/wp-settings.php#L377) [programmatically](https://developer.wordpress.org/reference/functions/wp_set_internal_encoding/).

PR #34, opened 2019-08-03
Adds Illuminate\Cache support

kellymears:
Adds support for Illuminate Cache to Acorn core.

### DIFF
--- a/config/cache.php
+++ b/config/cache.php
@@ -1,0 +1,104 @@
+<?php
+
+use function Roots\env;
+use Illuminate\Support\Str;
+
+return [
+
+    /*
+    |--------------------------------------------------------------------------
+    | Default Cache Store
+    |--------------------------------------------------------------------------
+    |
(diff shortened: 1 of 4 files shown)


config/cache.php
```diff
@@ -0,0 +1,104 @@
...
+
+        'database' => [
+            'driver' => 'mysql',
+            'table' => 'cache',
+            'connection' => null,
+        ],
+
+        'file' => [
+            'driver' => 'file',
+            'path' => get_theme_file_path('storage/framework/cache/data'),
```
QWp6t: We should probably use [`\Roots\storage_path('framework/cache/data')`](https://github.com/roots/acorn/blob/master/src/helpers.php#L253-L265) instead imo.

Log1x:
are we adding `config/cache.php` to sage 10?

config/cache.php
Log1x: updated

PR #36, opened 2019-09-01
log1x/composer-data

Log1x:
- feat(view): Add view composer `$data` variable 
- feat(view): Rename `getData()` to `merge()`
- chore(view): code style
- chore(view): docblocks

### DIFF
--- a/src/Acorn/View/Composer.php
+++ b/src/Acorn/View/Composer.php
@@ -68,7 +68,7 @@ abstract class Composer
     {
         return array_merge(
             $this->with(),
-            $this->data->toArray(),
+            $this->view->getData(),
             $this->override()
         );
     }
--- a/src/Acorn/View/Composer.php
+++ b/src/Acorn/View/Composer.php
@@ -3,15 +3,31 @@
 namespace Roots\Acorn\View;
 
 use Illuminate\Support\Str;
+use Illuminate\Support\Fluent;
 use Illuminate\View\View;
 
 abstract class Composer
 {
-    /** @var string[] List of views to receive data by this composer */
+    /**
+     * List of views to receive data by this composer
+     *
+     * @var string[]
+     */
     protected static $views;
 
-    /** @var \Illuminate\View\View Current view */
+    /**
+     * Current view
+     *
+     * @var \Illuminate\View\View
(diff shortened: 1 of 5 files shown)


src/Acorn/View/Composer.php
```diff
@@ -32,26 +48,27 @@ public static function views()
...
      *
      * @return array
      */
-    protected function getData()
+    protected function merge()
     {
         return array_merge(
             $this->with(),
-            $this->view->getData(),
+            $this->data->toArray(),
```
QWp6t: I think we should leave this as `$this->view->getData()`. The idea is that the view data remains untouched except what is being returned by `with()` and `override()` methods. By using `$this->data` here, it means that modifying `$this->data` within a view composer, you're directly modifying the data that's getting passed back to the view. It defeats the purpose of having separate `with()` and `override()` methods. Maybe that's desirable, I don't know; we should discuss it. It should be a separate issue/PR imo.

src/Acorn/View/Composers/Concerns/Cacheable.php
```diff
@@ -131,7 +131,7 @@ protected function getData()
 
         return array_merge(
             $with,
-            $this->view->getData(),
+            $this->data->toArray(),
```
QWp6t: Same as above.

src/Acorn/View/Composers/Debugger.php
```diff
@@ -20,13 +31,14 @@ public function compose($view)
...
-        
+
         $data = array_map(function ($value) {
             if (is_object($value)) {
                 return get_class($value);
             }
+
             return $value;
-        }, $view->getData());
+        }, $view->merge());
```
QWp6t: There is no `$view->merge()` method. This should be `$view->getData()`.

src/Acorn/View/Composers/Concerns/Arrayable.php
```diff
@@ -7,8 +7,18 @@
 
 trait Arrayable
 {
+    /**
+     * Ignored Methods
+     *
+     * @var array
```
QWp6t: `@var string[]`

PR #39, opened 2019-09-19
Add replaceSlugs filter for ComposerMakeCommand

Log1x:


src/Acorn/Console/Commands/ComposerMakeCommand.php
```diff
@@ -65,19 +67,36 @@ protected function buildClass($name)
...
      * @param  string  $stub
      * @param  array   $views
-     * @return string
+     * @return $this
      */
     protected function replaceViews($stub, $views)
     {
         $views = implode("',\n        '", $views);
-        return str_replace('DummyViews', empty($views) ? '//' : "'{$views}'", $stub);
+        $stub = str_replace('DummyViews', empty($views) ? '//' : "'{$views}'", $stub);
```
QWp6t: so what's actually being done with `$stub`?

src/Acorn/Console/Commands/ComposerMakeCommand.php
```diff
@@ -65,19 +67,36 @@ protected function buildClass($name)
...
+    /**
+     * Replace the slugs for the given stub.
+     *
+     * @param  string  $stub
+     * @param  string  $name
+     * @return $this
+     */
+    protected function replaceSlugs($stub, $name)
+    {
+        $stub = str_replace('DummySlug', Str::slug($name), $stub);
```
QWp6t: What's being done with `$stub`? And where is `DummySlug` set?

Log1x:
i sleepy i come fix & explain tmrw

Log1x:
ok I changed them back to returning strings. not sure if we wanted to make a separate method for running our "filters" (replacing views, slugs) – as far as where `DummySlug` is being used atm– in Composer but likely will replace a few use-cases in Acorn shortly too.

QWp6t:
What I mean is, there isn't a `DummySlug` anywhere in [composer.stub](https://github.com/roots/acorn/blob/master/src/Acorn/Console/Commands/stubs/composer.stub), so it seems as though that method isn't actually doing anything at the moment, right?

Log1x:
Actually, I'm an idiot and confused myself. We should just move these to [GeneratorCommand.php](https://github.com/roots/acorn/blob/master/src/Acorn/Console/Commands/GeneratorCommand.php)... I'm closing this and will redo.

PR #42, opened 2019-10-29
Update to Illuminate 6.3, add exception handling, code clean up

Log1x:
* feat(exceptions): Rewrite existing exception handler
* feat(exceptions): Add support for `filp/whoops`
* feat(exceptions): Add a Whoops handler for the WP Rest API and AJAX
* feat(exceptions): Add `WP`, `WP_Query`, and `WP_Post` to the Whoops `PrettyPageHandler`
* chore(deps): Update illuminate packages to 6.3
* chore(docblocks): Add missing docblocks
* chore(cs): Clean up various inconsistencies/smells in code
* chore(deps): Remove unused `illuminate/database` package and config
* chore(deps): Add `filp/whoops` to `require-dev` and `suggest`
* chore(deps): Add missing logging config
* chore(readme): Add contributing and community sections to README
* chore(package): Add Log1x as an author
* chore(package): Bump minimum PHP version to 7.2
* chore(circleci): Drop support for PHP 7.1
* chore(cs): Lint codebase


composer.json
```diff
@@ -31,29 +35,31 @@
...
   "require-dev": {
     "filp/whoops": "^2.5",
     "mikey179/vfsstream": "^1.6",
-    "phpunit/phpunit": "^7.5|^8.0",
+    "phpunit/phpunit": "^8.0",
     "roave/security-advisories": "dev-master",
     "squizlabs/php_codesniffer": "^3.3"
   },
+  "suggest": {
+    "filp/whoops": "Required for friendly error pages in development (^2.5)."
```
LeoColomb: `filp/whoops` is already required for dev, right? This suggestion will never be printed... 😉 
Oh and Laravel 6+ use https://github.com/facade/ignition, maybe a switch here too?

Log1x: filp/whoops will get printed when requiring Acorn in projects (e.g. Sage) and is required in those to initialize (a PR will be put out for Sage once this is merged).

Also we are looking at Ignition, but it will take some time before it is implemented (I have to rewrite most of the implementation as it hardcodes `illuminate/foundation`)

config/app.php
```diff
@@ -4,6 +4,58 @@
...
+    | Application Name
+    |--------------------------------------------------------------------------
+    |
+    | This value is the name of your application. This value is used when the
+    | framework needs to place the application's name in a notification or
+    | any other location as required by the application or its packages.
+    |
+    */
+
+    'name' => env('APP_NAME', wp_get_theme()->get('Name')),
```
QWp6t: This should grab the blog name by default, not the theme name lol

config/app.php
```diff
@@ -4,6 +4,58 @@
...
+    | Application Environment
+    |--------------------------------------------------------------------------
+    |
+    | This value determines the "environment" your application is currently
+    | running in. This may determine how you prefer to configure various
+    | services your application utilizes. Set this in your ".env" file.
+    |
+    */
+
+    'env' => env('WP_ENV', 'production'),
```
QWp6t: This doesn't account for people defining `WP_ENV`, such as if someone has put `define('WP_ENV', 'staging')` in wp-config.php.

It should probably be
```
 => defined('WP_ENV') ? WP_ENV : env('WP_ENV', 'production'),
```

config/app.php
```diff
@@ -4,6 +4,58 @@
...
+    | Application Debug Mode
+    |--------------------------------------------------------------------------
+    |
+    | When your application is in debug mode, detailed error messages with
+    | stack traces will be shown on every error that occurs within your
+    | application. If disabled, a simple generic error page is shown.
+    |
+    */
+
+    'debug' => env('WP_DEBUG', false),
```
QWp6t: This should just be `WP_DEBUG` since WordPress will always define it.

config/app.php
```diff
@@ -15,7 +67,7 @@
     |
     */
 
-    'preflight' => false,
+    'preflight' => env('WP_ENV', 'production') !== 'production',
```
QWp6t: We don't want to assume that only the production environment shouldn't be running preflight. It goes against environment parity. Instead we assume all environments are equivalent with production by default. But even if we were to keep this, it has the same issue as before with using `env()`.

src/Acorn/Bootstrap/HandleExceptions.php
```diff
@@ -0,0 +1,180 @@
...
+    protected $app;
+
+    /**
+     * Bootstrap the given application.
+     *
+     * @param  \Roots\Acorn\Application  $app
+     * @return void
+     */
+    public function bootstrap(Application $app)
+    {
```
QWp6t: This should return early if `WP_DEBUG` is false.

I'm wondering if we should also check [`is_readable(WP_CONTENT_DIR . '/fatal-error-handler.php')`](https://github.com/WordPress/WordPress/blob/master/wp-includes/error-protection.php#L88).

src/Acorn/Bootstrap/SageFeatures.php
```diff
@@ -6,35 +6,39 @@
 
 class SageFeatures
 {
+    /**
+     * Sage Features
+     *
+     * @var array
+     */
+    protected $features = [
+        'sage' => \Roots\Acorn\Sage\SageServiceProvider::class,
```
QWp6t: `sage` should always get loaded if either `assets` or `blade` are loaded.

src/Acorn/Concerns/Application.php
```diff
@@ -537,7 +537,7 @@ public function isBooted()
     protected function fireAppCallbacks(array $callbacks)
     {
         foreach ($callbacks as $callback) {
-            call_user_func($callback, $this);
+            $callback($this);
```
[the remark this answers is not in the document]
Log1x: This syntax will work just fine for the `[$object, 'method']` syntax. It is `call_user_func_array()` that still has purpose in later versions of PHP. You can see in Illuminate 6.0 they changed all of their `call_user_func()`'s to call the method with the variable instead due to performance reasons as well as it just being cleaner syntax.

src/Acorn/Bootstrap/HandleExceptions.php
Log1x: Checking for `fatal-error-handler.php` sounds sane. I will add that. I'm also considering since we want all of the exception handler to be opt-in, if I should just create a new Acorn-specific value in `app.php` to enable/disable it alongside `WP_DEBUG`?

src/Acorn/Bootstrap/HandleExceptions.php
```diff
@@ -0,0 +1,180 @@
...
+
+        error_reporting(-1 & ~E_USER_NOTICE);
+
+        set_error_handler([$this, 'handleError']);
+        set_exception_handler([$this, 'handleException']);
+        register_shutdown_function([$this, 'handleShutdown']);
+
+        if (! $app->environment('development')) {
+            ini_set('display_errors', 'Off');
+        }
```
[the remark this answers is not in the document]
Log1x: Will do.

src/Acorn/Bootstrap/HandleExceptions.php
```diff
@@ -0,0 +1,180 @@
...
+     * @param  \Roots\Acorn\Application  $app
+     * @return void
+     */
+    public function bootstrap(Application $app)
+    {
+        self::$reservedMemory = str_repeat('x', 10240);
+
+        $this->app = $app;
+
+        error_reporting(-1 & ~E_USER_NOTICE);
```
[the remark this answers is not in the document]
Log1x: The reason for `~E_USER_NOTICE` is due to Whoops catching `trigger_error()` and `_doing_it_wrong()` as exceptions – there are many instances of this happening in WordPress core alone causing an instant halt of the site.

An example of this happening is using Bedrock and going to `http://example.com/wp-admin` – WP triggers a `trigger_error()` when it runs an improper query to WordPress via the 404/canonical redirection function and throws an error before actually redirecting to `/wp/wp-admin`.

config/app.php
Log1x: Let's just stick to `false`.

src/Acorn/Bootstrap/HandleExceptions.php
QWp6t: There's `WP_DEBUG_DISPLAY` that WordPress uses. We could require that to be true. Or we could create our own config option that derives its default value from it.

config/logging.php
```diff
@@ -45,13 +45,13 @@
...
             'driver' => 'single',
-            'path' => storage_path('logs/sage.log'),
+            'path' => storage_path('logs/app.log'),
             'level' => 'debug',
         ],
 
         'daily' => [
             'driver' => 'daily',
-            'path' => storage_path('logs/sage.log'),
+            'path' => storage_path('logs/app.log'),
```
QWp6t: lol good call

src/Acorn/Bootstrap/SageFeatures.php
Log1x: fixed & cleaned up

src/Acorn/Bootstrap/HandleExceptions.php
Log1x: Lets stick with `WP_DEBUG` which is our new default for `app.debug` – the handler will now bail if `debug` is set to `false` or a fatal error handler exists **unless** using acorn cli.

src/Acorn/Bootstrap/HandleExceptions.php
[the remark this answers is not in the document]
Log1x: this is all behind `app.debug` now

PR #43, opened 2019-11-15
feature(asset): Add get() method allowing you to return the asset value

Log1x:


src/Acorn/Assets/Asset.php
```diff
@@ -57,9 +57,20 @@ public function contents(): string
...
     }
 
+    /** {@inheritdoc} */
+    public function get(): array
+    {
+        if (! $this->exists()) {
+            return false;
+        }
+
+        return require_once($this->path());
```
QWp6t: RRRREEEEEEEEEEEEEEEEEEEE

`require_once $this>path();`

But I also think we should use `include`.

PR #44, opened 2019-12-05
Console Features/Improvements

Log1x:
* feature(console): implement custom summary command for displaying console help
* feature(console): add clearLine helper method
* feature(console): add exec helper method for executing processes
* feature(console): add task helper for creating styled tasks
* chore(console): remove unused schedule method/namespaces
* chore(docblock): clean up docblocks

### DIFF
--- a/src/Acorn/Console/Commands/Command.php
+++ b/src/Acorn/Console/Commands/Command.php
@@ -27,7 +27,7 @@ abstract class Command extends CommandBase
      * Write a string in a title box.
      *
      * @param  string $title
-     * @return void
+     * @return \Roots\Acorn\Console\Commands\Command
      */
     public function title($title)
     {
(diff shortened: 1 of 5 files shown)


src/Acorn/Console/Commands/Command.php
```diff
@@ -2,25 +2,19 @@
...
      * @var \Roots\Acorn\Application
      */
     protected $app;
 
-    /**
-     * {@inheritdoc}
-     */
-    public function schedule(Schedule $schedule)
-    {
-    }
```
QWp6t: Is there a reason we had this in the first place? Why is it being removed?

Log1x: I'm 99% sure it was required initially to extend `Command` but it appears to of [been removed](https://github.com/illuminate/console/blob/master/Command.php) in 6.x although I'm not sure which commit. In any case, everything is fine without it.

PR #47, opened 2020-01-20
feature(asset): add toJson() method for reading asset manifests, etc.

Log1x:
Necessary little helper for stuff that require you to read the manifest like Encore's `entrypoints.json`, etc.

Will also clean up the Sage 10 implementation of [https://github.com/roots/palette-webpack-plugin](palette-webpack-plugin)

### DIFF
--- a/src/Acorn/Assets/Asset.php
+++ b/src/Acorn/Assets/Asset.php
@@ -75,6 +75,20 @@ class Asset implements AssetContract
         return include $this->path();
     }
 
+    /**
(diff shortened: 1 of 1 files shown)


QWp6t:
I think a better implementation would be to use an `AssetFactory` that generates `Asset` objects based on asset type (such as file extension), some of which can be `Jsonable`, otherwise you're adding `Asset::toJson()` to all assets, the majority of which are likely to not be json files.

PR #60, opened 2020-04-21
Potential fix for broken view autoloading

alwaysblank:
In previous releases, you could forgo adding a `$views` prop if your namespacing and class names reflected the structure of your views. It looks (??) like a namespacing change at some point broke this, as I wasn't able to get it to work on a recent project.

This commit fixes the issue, but is likely just a hack; consider this a prompt for a better solution if mine is problematic.

### DIFF
--- a/src/Acorn/View/Composer.php
+++ b/src/Acorn/View/Composer.php
@@ -41,6 +41,9 @@ abstract class Composer
         }
 
         $view = array_slice(explode('\\', static::class), 2);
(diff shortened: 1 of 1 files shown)


QWp6t:
I didn't notice this PR at the time, but I resolved this in https://github.com/roots/acorn/commit/daa4fa90e4502e5919832347fb8a690d2e2776b1

Sorry about that. 😬

PR #67, opened 2020-06-03
fix(helpers): view() should support file paths

oxyc:
Currently the [`template_include` filter](https://github.com/roots/acorn/blob/master/src/Acorn/Sage/Concerns/FiltersTemplates.php#L40) sets the view to the template path as a fallback when no view exists. 

This is not actually supported by `Roots/view()` but this patch should fix it although I'm not sure it's the correct way.

### DIFF
--- a/src/helpers.php
+++ b/src/helpers.php
@@ -404,7 +404,8 @@ function view($view = null, $data = [], $mergeData = [])
         return $factory;
     }
 
-    return $factory->make($view, $data, $mergeData);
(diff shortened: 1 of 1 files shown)


QWp6t:
Your implementation looks good to me. That's exactly how I used to have it.

https://github.com/roots/acorn/blob/fc4b1457199585094d2f35d2cc83ee3d172a08b9/src/helpers.php#L93-L95

Thanks!

PR #70, opened 2020-06-22
Small bugfix for unit test

w-sch:
The CI is red because I expected a path that is different in CI.

* Remove "acorn" from test string as it is not present in CI pipeline


w-sch:
Can we activate the CI tests for PRs? Like that I see if the change works as expected before approving the PR.

Log1x:
> Can we activate the CI tests for PRs? Like that I see if the change works as expected before approving the PR.

CI should trigger on PR's now. Sorry about that.

PR #74, opened 2020-09-29
Acorn 2.x (Laravel 8.x)

QWp6t:
This is a fairly significant restructuring of things.

I haven't extensively tested everything, but I did make sure a sage 10 site loads without any errors and the wp-cli commands are working.

Tests are going to be failing because they need to be refactored. I can handle that this weekend.

(I'll document all of the changes I made later.)

PR #75, opened 2020-10-19
Support for Acorn placed in require section of composer.json in root of Bedrock page with Sage theme

freezy-sk:
Support for Acorn placed in require section of composer.json in root of Bedrock page with Sage theme.
This also allows us to get rid of vendor directory from Sage theme folder.

Example of autoload section of composer.json in page root.
```
"autoload": {
    "psr-4": {
        "App\\": "web/app/themes/your-theme-name/app/"
    }
}
```

### DIFF
--- a/src/Acorn/ComposerScripts.php
+++ b/src/Acorn/ComposerScripts.php
@@ -2,6 +2,7 @@
 
 namespace Roots\Acorn;
 
+use Composer\Autoload\ClassLoader;
 use Composer\Script\Event;
 use Roots\Acorn\Application;
 use Roots\Acorn\Filesystem\Filesystem;
@@ -17,10 +18,11 @@ class ComposerScripts
      */
     public static function postInstall(Event $event)
     {
-        require_once $event->getComposer()->getConfig()->get('vendor-dir') . '/autoload.php';
+        $loader = require $event->getComposer()->getConfig()->get('vendor-dir') . '/autoload.php';
+        $basePath = static::getAppBasePath($loader);
 
-        static::clearCompiled();
-        static::buildManifest();
+        static::clearCompiled($basePath);
+        static::buildManifest($basePath);
     }
 
     /**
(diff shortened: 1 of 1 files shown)


AmphiBee:
great pull request :)

QWp6t:
I hesitate to merge this because the `App\\` namespace should not be assumed. We're basically trading the assumption that the base path is the theme directory for the assumption that the application namespace is `App\\`.

On the other hand, I wouldn't be entirely opposed to merging it for now until Acorn 2.x is finished.

Have you looked into whether there's a namespace-agnostic way of accomplishing this?

For what it's worth, we're going to have much broader support for this in the upcoming version of Acorn.

1. There are [several ways to specify the base path](https://github.com/roots/acorn/blob/laravel8/src/Roots/Acorn/Bootloader.php#L167-L169). (you can do this in [Acorn 1.x as well](https://github.com/roots/acorn/blob/master/src/Acorn/Bootloader.php#L130-L132))
2. `app`, `config`, `storage`, and `resources` are all [filterable](https://github.com/roots/acorn/blob/laravel8/src/Roots/Acorn/Bootloader.php#L183).
3. All other paths can be modified using [`$app->usePaths()`](https://github.com/roots/acorn/blob/laravel8/src/Roots/Acorn/Application.php#L91-L137)
4. We're using [`wp acorn`](https://github.com/roots/acorn/blob/laravel8/src/Roots/Acorn/Console/Console.php) for our [composer scripts](https://github.com/roots/acorn/blob/laravel8/src/Roots/Acorn/ComposerScripts.php).

freezy-sk:
@QWp6t I chose `App\` namespace after looking at how it's all scaffolded in Sage 9 vs 10 and changes introduced in Acorn 2.0 (I see many core Laravel classes added to Acorn 2.0)

This theme application base path guessing can be changed to another ways like config in composer's extra or env var like it's used for Acorn's base path.
Could you describe what are the plans for future of Acorn?
If it will be used only as support for themes, I will choose that default of App\ namespace because Acorn 2.0 is so tightly coupled with Laravel.
If it should be used as base for plugins, this path guessing must be done another way.

QWp6t:
Thanks for the PR. 👍

We have taken a different approach to address this, and it has been implemented in the main branch as part of our Acorn 2.x updates (#74).

We will tag a release sometime soon.

PR #77, opened 2020-11-17
Add Wordpress filters support

plumthedev:
For a while I use a solution for better handling WordPress filters and actions.

Every filter class is stored in configuration as class name string and it is build on application bootstrap by container. By this filter class can have easily access to dependency injection on construct.
Every filter instance must have handle method which is `handle` by default.
Count of arguments for handle method is automatically calculated by reflection mechanism.
Filter priority is object property with 10 value by default.
Filter is applied by `\Roots\add_filters` and removed by `Roots\remove_filters`

All is contracted and fully typed, in all my cases this solution was doing great.


stefanfisk:
How would this work in situations like this, one filter class per hooks? https://github.com/roots/acorn/blob/fa60ba1551dc96400b10aaa7f396a2fffd62e5c6/src/Acorn/Sage/SageServiceProvider.php#L53

plumthedev:
it is good question - as you can see the same function is used in `Roots\Acorn\Support\Filter::apply()` and return type of `Roots\Acorn\Support\Filter::getTag()` is iterable so this filters supports this situation as well :)  
  
https://github.com/plumthedev/acorn/blob/feature/add-wordpress-filters-support/src/Acorn/Support/Filter.php#L44  
  
https://github.com/plumthedev/acorn/blob/feature/add-wordpress-filters-support/src/Acorn/Support/Filter.php#L143

plumthedev:
Okay, now I see is default tag is type of string not iterable, I have to change it to array of strings and that will be fair enough. 

I try to keep Acorn ecosystem as it is because it solves problems like one handle with multiple hooks out of the box 😁

stefanfisk:
Nice! And if one wants to apply a filter on different priorities for different tags I guess that you could just override `apply()` and `remove()` and do whatever you need to do, correct?

For example, I've got a filter that adds `&shy;` to words based on a dictionary, and it hooks in all over the place and sometimes the priority is important.

plumthedev:
This is very edge case example but your suggestion is absolutely enough.  
  
I used this filters in real-life projects and it was doing great job. Everything is clarify and can collect more business logic in one place, without several thousand lines of code in `functions.php` or something like that. With dependency injection this tool is very powerful in my opinion.

stefanfisk:
I in no way meant to critique - it looks great! I was only asking because I'm super eager to use this pattern instead of what I am currently doing, and I wanted to get a clear picture of how to best go about handling the more complex cases.

plumthedev:
Of course :smile: I appreciate your words on this PR because you found small bug with iterable tag.

QWp6t:
Thanks for the PR.

After discussing this internally, we've decided to pass on this one. Although I like the goals of what you're doing, and I think this would be great as a separate library from Acorn that gets registered as a service provider.

PR #83, opened 2020-12-23
Bind paths in container

### LINKED ISSUE #82: public_path() breaks because paths are not bound
* [x] I've read the [guidelines for Contributing to Roots Projects](https://github.com/roots/.github/blob/master/CONTRIBUTING.md)
* [x] This request isn't a duplicate of an existing issue
* [x] I've read the [docs](https://roots.io/documentation/) and followed them (if applicable)
* [x] This is not a personal support request that should be posted on the [Roots Discourse](https://discourse.roots.io/) community

## Description

`public_path()` throws because paths (`path.*`) are not bound in the app container: https://github.com/roots/acorn/blob/master/src/helpers.php#L290

This might also cause problems with other Laravel code that expects the paths to be bound.

## Steps to reproduce

Call `public_path()`.

**Expected behavior:**

It should return the path.

**Actual behavior:**

It throws due to the missing binding.

**Reproduces how often:**

Every time.

## Versions

`Roots\Acorn:1.1.0`

## Additional information

N/A.

stefanfisk:
Fixes #82.

### DIFF
--- a/src/Acorn/Application.php
+++ b/src/Acorn/Application.php
@@ -67,11 +67,44 @@ class Application extends Container implements ApplicationContract
     public function __construct($basePath = null)
     {
         if ($basePath) {
-            $this->basePath = $basePath;
+            $this->setBasePath($basePath);
         }
 
         $this->registerContainerBindings();
         $this->bootstrapContainer();
+    }
+
+    /**
+     * Set the base path for the application.
+     *
+     * @param  string $basePath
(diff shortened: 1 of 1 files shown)


QWp6t:
Thanks for the PR. 👍

This change has been implemented in the main branch as part of our Acorn 2.x updates (#74).

We will tag a release sometime soon.

PR #85, opened 2021-01-28
Change null to empty string

ouun:
`comments_template` filter expects a string but `null` is given. Small fix also for other `comments_template` filters with lower priority.
Also see: https://github.com/generoi/sage-woocommerce/pull/6


QWp6t:
We can merge this if you can update it to latest version and return `$file` instead of an empty string.

If you'd rather me handle it, let me know.

ouun:
@QWp6t thanks for your reply. I changed it to `$file`.  However I am afraid I do not understand what you mean with "update it to latest version".

Jamiewarb:
Merge roots:main into your branch @ouun, to resolve the conflicts and update it with the latest version from upstream

ouun:
Thanks @Jamiewarb. I might have overcomplicated it, but latest changes are merged now.

PR #95, opened 2021-03-02
Fix commands registration

mikeyzm:
Follow-up to #93


Log1x:
fixed this regression in master. sorry about that.

PR #101, opened 2021-04-12
feat(sage): add privacypolicy and embed filters

huubl:
Add missing `privacypolicy_template_hierarchy` and `embed_template_hierarchy` to filter.

I found that `privacy-policy.blade.php` didn't load, this fixes it.

Reference:
https://github.com/WordPress/WordPress/blob/master/wp-includes/template.php#L37-L62


Log1x (approved):

QWp6t:
Thanks for the PR!

PR #105, opened 2021-05-24
fix(config-view): use commented example directive

pxlrbt:
Apparently the `AssetDirective` was renamed/moved to  `Roots\Acorn\Assets\View\BladeDirective`.
With the wrong directive defined, Acorn 2.0.0-alpha.0 doesn't register View Composers and throws an error for a missing var.

### DIFF
--- a/config/view.php
+++ b/config/view.php
@@ -79,6 +79,6 @@ return [
     */
 
     'directives' => [
-        'asset'  => Roots\Acorn\Assets\View\BladeDirective::class,
+        // 'foo'  => App\View\FooDirective::class,
     ],
 ];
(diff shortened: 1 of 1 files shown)


QWp6t:
It's [loaded automatically by the `AssetsServiceProvider`](https://github.com/roots/acorn/blob/main/src/Roots/Acorn/Assets/AssetsServiceProvider.php#L36)

But this is still an issue that should be corrected, so if you want to change that line to [match what we have in Sage](https://github.com/roots/sage/blob/master/config/view.php#L82), I can merge this.

pxlrbt:
Thanks. I didn't see that this is autoloaded now. I updated the config to include the commented examples line,

PR #107, opened 2021-05-26
fix: error handling only when debugging enabled

Log1x:
- fix(exceptions): Do not initialize the error handler unless debug is `true`


Log1x:
related: https://github.com/roots/sage/issues/2656

Issue #109, opened 2021-05-27
Symfony Error Handler broken


pxlrbt:
* [ ] I've read the [guidelines for Contributing to Roots Projects](https://github.com/roots/.github/blob/master/CONTRIBUTING.md)
* [x] This request isn't a duplicate of an existing issue
* [ ] I've read the [docs](https://roots.io/documentation/) and followed them (if applicable)
* [x] This is not a personal support request that should be posted on the [Roots Discourse](https://discourse.roots.io/) community

## Description

The Symfony error handler is never rendering which leads to blank pages on errors when `WP_ENV` is not set to `development`

## Steps to reproduce

1. Set `WP_ENV` to "local"
2. Trigger an error/exception

**Expected behavior:** Symfony error page

**Actual behavior:** Blank page

**Reproduces how often:** 100%

## Versions
2.0.1

Gandhi11:
I have the same problem. You only need to install your composer dependencies without the dev dependencies and you will have the problem.

1 - The error needs to be send() to receive the proper html error code.
In the `src/Roots/Acorn/Bootstrap/HandleExceptions.php` file, the method ->send() needs to be append after the render in the renderHttpResponse method. Like so.
```
$this->getExceptionHandler()->render('', $e)->send();
```

2 - The error is not render properly in the page.
In the `/src/Roots/Acorn/Exceptions/Handler.php` file, the render need to be echoed.
```
    public function render($request, Throwable $e)
    {
        try {
            echo app()->environment('development') && class_exists(Whoops::class)
                        ? $this->renderExceptionWithWhoops($e)
                        : $this->renderExceptionWithSymfony($e, app()->environment('development'));
        } catch (Exception $e) {
            echo $this->renderExceptionWithSymfony($e, app()->environment('development'));
        }
    }
```

pxlrbt:
Hey @Gandhi11,
I already opened a PR here: https://github.com/roots/acorn/pull/11
I only changed what you did in your 2. What's the `->send()` for? Will this prevent previous output? I couldn't figure how to prevent that. Maybe you could comment on that PR?

Gandhi11:
If the `->send()` is not used. The page still return a valid 200 OK even if there is a server error.

pxlrbt:
That doesn't work for me:

Currently `$this->getExceptionHandler()->render('', $e)` returns an instance of `Whoops/Run` - which doens't have a `->send` method or string.

With adjustment 2 the return value is `void`.

[...]

PR #110, opened 2021-05-27
fix(exception-handler): Symfony error handler not rendering

pxlrbt:
This PR fixes [Issue 109](https://github.com/roots/acorn/issues/109) which returns a blank page instead of the symfony error handler.

The issue is: While Whoops writes directly to the output, the symfony handler is only returning the HTML and thus not rendering at all. I reverted the custom Whoops handler and added an `echo` statement.


pxlrbt:
Just made some more adjustments:
- `Acorn/Exceptions/Handler::render` now returns a `Symfony\Component\HttpFoundation\Response` as stated in the annotation. This will also send the right HTTP Code (500) now. 
- I added a `ob_clean` inside `Roots\Acorn\Bootstrap\HandleExceptions` to prevent any data from leaking the Exception Handler.

PR #122, opened 2021-10-30
Next: switch to pest, enable globals, fix errors/exception handling.

### LINKED ISSUE #87: Error Handler Broken
It looks like master is already on 2.0 now, which doesn't have this issue, but wanted to raise it in case there is to be a maintenance release of the 1.x series, and to let others know about it.

#58 was to add functionality to ignore certain classes of error, adding the following code:
https://github.com/roots/acorn/blob/fa60ba1551dc96400b10aaa7f396a2fffd62e5c6/src/Acorn/Bootstrap/HandleExceptions.php#L186-L197

But from what I can see, this code doesn't work. It's checking if `ErrorException` is an instance of `'E_USER_DEPRECATED'`, which I don't believe will ever work. Something like:
```
$e->getSeverity() === constant($type)
```
is what I can come up with to get this to work as designed.

My workaround in a Sage 10 theme with Acorn v1.1.0 is to check if the Acorn error handler is the one registered, and if so, replace it with one that unconditionally throws an ErrorException

```
add_action('after_setup_theme', function () {
    $previous_handler = set_error_handler(null);

    // Check if the Acorn error handler was actually the one we just removed or not
    if (!empty($previous_handler[0]) && $previous_handler[0] instanceof \Roots\Acorn\Bootstrap\HandleExceptions) {
        set_error_handler(function ($level, $message, $file = '', $line = 0, $context = []) {
            throw new ErrorException($message, 0, $level, $file, $line);
        });
    } else {
        set_error_handler($previous_handler);
    }
}, 6);
```

### LINKED ISSUE #100: Error handler: incorrect bitwise comparison
* [x] I've read the [guidelines for Contributing to Roots Projects](https://github.com/roots/.github/blob/master/CONTRIBUTING.md)
* [x] This request isn't a duplicate of an existing issue
* [x] I've read the [docs](https://roots.io/documentation/) and followed them (if applicable)
* [x] This is not a personal support request that should be posted on the [Roots Discourse](https://discourse.roots.io/) community

## Description

Note: `acorn` put the code of a previously deprecated package (on packagist) into its own repo.

There is a bug with the bitwise operation used in the condition to check for `error_reporting` as it doesn't work for disabled error_reporting or `error_reporting` without lower reporting levels:
https://github.com/roots/acorn/blob/a2ef0a5fe7a4b4d692838529a1b67c48245f0cbd/src/Illuminate/Foundation/Bootstrap/HandleExceptions.php#L68

with `error_reporting` set to `0` or `E_ERROR | E_WARNING | E_PARSE` in order to skip `E_NOTICE` PHP errors, 
the error will still be thrown, shown and destroy the page markup at that point with injected "rich" error HTML.

## Steps to reproduce

1. Configure for lower or none `error_reporting`, e.g. for staging in `config/environments/staging.php`
2. Trigger a PHP `E_NOTICE` message, e.g. an `Undefined index` error by accessing an undefined index of a PHP array:
Note that these lower level PHP errors still cause a "rich" HTML error message, which shouldn't happen.

**Expected behavior:**
PHP errors with levels to be ignored.

**Actual behavior:**
PHP errors with levels to be ignored are still thrown.

**Reproduces how often:**
Always.

## Versions
Bedrock:
````
### 1.15.3: 2021-03-12
* chore(deps): Bump roots/wordpress from 5.6.2 to 5.7 ([#578](https://github.com/roots/bedrock/pull/578))
````

## Additional information
Related: https://github.com/roots/acorn/issues/87

### LINKED ISSUE #109: Symfony Error Handler broken

### LINKED ISSUE #111: [Discussion] Exception handler behaviour
The exception handler currently has some strange behavior:

1. `WP_DEBUG = true` && `WP_ENV = 'production'` returns a blank page due to [issues with Symfony handler](https://github.com/roots/acorn/issues/109). Also the Symfony handler is pretty useless for debugging. Not sure what's the best way to handle this. 
2. `WP_DEBUG = false` && `WP_ENV = 'production'` doesn't use an Exception Handler and prints PHP warnings/errors depending on WPs `error_reporting` setting. It returns more information than (1) which seems weird, as (1) is "debug mode".
3. `WP_DEBUG = false` && `WP_ENV = 'development'`: Is fine as it is.
4. `WP_DEBUG = true` && `WP_ENV = 'development'` : Returns whoops which is also fine

Generally I think `WP_DEBUG_DISPLAY` would have been a better indicator whether to show an exception handler as it's already a standard within WordPress. But that probably would cause some issue with Sage?!

### LINKED ISSUE #113: Error handler throws Exception for notices
* [ ] I've read the [guidelines for Contributing to Roots Projects](https://github.com/roots/.github/blob/master/CONTRIBUTING.md)
* [x] This request isn't a duplicate of an existing issue
* [x] I've read the [docs](https://roots.io/documentation/) and followed them (if applicable)
* [x] This is not a personal support request that should be posted on the [Roots Discourse](https://discourse.roots.io/) community

## Description

Similar to other issues with the error handler:
the Error handler is too strict on throwing exceptions, returning the Whoops page as a response breaks many plugins' functionalities.

## Steps to reproduce

1. Cause 'Undefined Index' notice
2. Get Error-page

**Expected behavior:**
Notice should be sent but code should not be interrupted

**Actual behavior:**
A Whoops-error page is returned and code is exited

**Reproduces how often:**
100%

## Versions
2.0.0.alpha

### PR DESCRIPTION
**Pull request by QWp6t**
Closes #87
Closes #100
Closes #109
Closes #110
Closes #111
Closes #112
Closes #113

### CODE REVIEW

File: composer.json [hunk 1]
```diff
@@ -50,7 +50,6 @@
     "illuminate/log": "^8.0",
     "illuminate/support": "^8.0",
     "illuminate/view": "^8.0",
-    "laravel-zero/foundation": "^8.0",
```
Kilbourne: Hi,
Can you explain me why you have chosen to have a copy of foundation in the source code vs using the one provided by laravel-zero?

retlehs: via @QWp6t: 

> Because it autoloads `helpers.php`, which contains a bunch of global functions, one (or more?) of which conflicts with global helper functions provided by WordPress, such as `__()`

Kilbourne: Please check this composer plugin [https://github.com/mcaskill/composer-plugin-exclude-files](https://github.com/mcaskill/composer-plugin-exclude-files) which was made for this case ( excluding laravel helpers from foundation).
So much better to have Foundation in vendor code..

QWp6t: I had considered that, but in my testing, it would require us to instruct users to add extras.exclude-from-files manually to their composer.json file, i.e., we can't just include that with Acorn.

This isn't the end of the world, but it does mean that we'd frequently get issues and bug reports from users who miss this step and can't figure out why there's a "function already defined" error on their site. Additionally, that plugin doesn't run if the user passes `--no-scripts` to `composer install`.

It seemed easier to continue to vendor Foundation for now.

But obviously I'm totally open to the idea of switching back to laravel-zero/foundation if we can come up with an elegant zero-config way of loading it without helpers.php.

Kilbourne: IMO just add it to composer.json of Sage and you have covered +90% of cases. 
For other uses, like i'm doing adding acorn to the project composer.json, you will need anyway to make some adjustments and use Sage's composer.json as reference to copy useful parts...
for `--no-scripts`, probably you mean ` --no-plugins` which seems an even rarer case..
As further solution could be to create a Roots fork of laravel-zero/Foundation with this little modification to the composer.json ( a github action could automate the process on each release)

QWp6t: We're [no longer going to have Acorn as a direct dependency of Sage](https://github.com/roots/sage/pull/2827). That's what prompted me to remove laravel-zero/foundation and re-add Foundation directly into the repo.

We will eventually set up some sort of automated solution, but for right now, I'm focusing my time on getting some better test coverage in Acorn and squashing bugs. When things settle down and I have some time, I'll revisit the idea of setting up our own repo to mirror Foundation with some modifications.

strarsis:
@QWp6t: Has this already been released (acorn) and required by Sage 10?
I just re-encountered this issue again with a recent Sage 10 theme.

Log1x:
what issue?

strarsis:
@Log1x: PHP notices and warnings causes a whoops error/exception that makes the whole page unusable.

Log1x:
this should only happen on development. we could probably use an `is_admin()` check though. thoughts @QWp6t ?

strarsis:
@Log1x: Good point. Though this also happens for me in production 😿 (blank page then, but corresponding error log entry).

Log1x:
is this primarily happening in admin?

strarsis:
No, it also happens on frontend. But most PHP notices/warnings (e.g. WooCommerce) happen in admin, so I notice them mostly there.

PR #141, opened 2021-11-20
Template hierarchy filter: Preserve existing paths

strarsis:
@retlehs: Instead of solely returning the blade template path locate result, 
append it to the array of existing files passed to the template hierarchy filter handler,
allowing Gutenberg Full Site Editing (FSE) to find and use block templates/template parts in the (Sage 10) theme.

This fix lets the Gutenberg Full Site Editing/FSE templates load in Gutenberg, 
while still allowing classic (blade) templates to be used (as for WooCommerce that currently doesn't fully use Gutenberg yet).

### Useful resources
- Workaround that short-circuits the filter instead (as a fix in a Sage 10 theme): roots/acorn#228

- Discussion about the reasons on why Gutenberg Full Site Editing/FSE can't find block templates/template parts without the fix:
https://discourse.roots.io/t/full-site-editing-fse-frontend-doesnt-load-template/21574/4?u=strarsis

- Related issue in the Sage 10 repository (that uses this `acorn` library in its PHP runtime): roots/acorn#228

### PSA
***Double-check whether `remove_theme_support('block-templates')` is absent from your theme setup, as otherwise FSE will not work.***


retlehs:
@strarsis thanks for the PR! i'm seeing some really odd behavior when trying to test FSE

could you share a repo/branch with me that has sage 10 with your FSE changes applied to it?

strarsis:
@retlehs: Sure: https://github.com/strarsis/acorn/tree/patch-1

strarsis:
@retlehs: I see, yes I can prepare something that uses this patch and also has FSE features so this can be tested.

strarsis:
@retlehs: Alright, this repository includes this patch, enables Gutenberg Full Site Editing/FSE and adds FSE block templates and block template parts as demonstration.

stefanfisk:
@strarsis did you forget the link, and if so is this the one https://github.com/strarsis/sage10-fse?

strarsis:
Yes! I forgot to link the repo.

strarsis:
@stefanfisk: The [repo](https://github.com/strarsis/sage10-fse) was updated and works now with latest WordPress, Gutenberg plugin and Sage 10 theme.
Discussion: https://discourse.roots.io/t/full-site-editing-fse-frontend-doesnt-load-template/21574/16?u=strarsis

ouun:
Any news on this? FSE support with Acorn/Sage would be really great.

oxyc:
I'm running it fine by overriding the SageServiceProvider and not [binding filters](https://github.com/roots/acorn/blob/4d902cc6fca9134b97a6b95ccdbe9ae05a1d1575/src/Roots/Acorn/Sage/SageServiceProvider.php#L31-L32), so that all templates are in FSE. Running it with WooCommerce

strarsis:
@oxyc: Would this be a better approach than modifying the `acorn` template filter as in this PR?
@ouun: If you have the time, could you try out the linked Sage 10 FSE theme example and see whether it works? It is Sage 10 with the patched `acorn` library.

oxyc:
To me it makes sense to not the bind the filters in the first place. I have no php templates in my theme since I opted in to FSE. I can still use blade in blocks

strarsis:
@oxyc: Speaking of unbinding or short-circuiting filters, I had to force the Blade-PHP template filter to skip when a PHP-file already exists. Without this additional adjustment, some part of the HTML page markup would be rendered twice (`<header>`), which wouldn't be noticeable visually, but cause unexpected behavior (Popup maker).

So ideally the Sage template filter can use Blade-PHP templates, fall back to block templates and finally usual PHP files.
The adjustments in this PR would allow for this behavior.

strarsis:
Side note/FWIW: fellow devs who still need PHP 7.x support, the `patch-v2` branch has this fix also applied, but on the highest possible `acorn` version [that still supported `PHP 7.3`](https://github.com/strarsis/acorn/blob/patch-v2/composer.json#L40): https://github.com/strarsis/acorn/tree/patch-v2

broskees:
Does this need more work to get into acorn core?

strarsis:
There is a modification by @r-chrzan which appears to improve the loading behavior (Blade over FSE template files):
https://github.com/strarsis/sage10-fse/issues/1#issuecomment-1480357953

palicko:
Hey guys, how does this PR looks like? You think it can be merged into Acorn core any time soon? 
Thanks!

strarsis:
> Hey guys, how does this PR looks like? You think it can be merged into Acorn core any time soon? 
> Thanks!

By the way, there is also another PR now that may handle hybrid theme situations better: https://github.com/roots/acorn/pull/314

Independent from the particular acorn fix, you may find this Sage 10 FSE example theme base interesting:
https://github.com/strarsis/sage10-fse

Log1x (approved):

strarsis:
Awesome! 👍

Issue #142, opened 2021-11-23
Wrong namespace for aliases in Application::registerCoreContainerAliases()


stefanfisk:
* [x] I've read the [guidelines for Contributing to Roots Projects](https://github.com/roots/.github/blob/master/CONTRIBUTING.md)
* [x] This request isn't a duplicate of an existing issue
* [x] I've read the [docs](https://roots.io/documentation/) and followed them (if applicable)
* [x] This is not a personal support request that should be posted on the [Roots Discourse](https://discourse.roots.io/) community

## Description

There's a couple of aliases with incorrect namespace in
https://github.com/roots/acorn/blob/9af5ffff189abd09436d4280df13c620675252a3/src/Roots/Acorn/Application.php#L274

Even when fixing them I was not able to make inject `Illuminate\Filesystem\Filesystem` after boot, but I didn't have time to debug that now.

QWp6t:
Yeah, my bad. I'm not sure at which point that got screwed up, but it should be resolved now.

Feel free to check out #143 if you need this right away, otherwise I'll review it more thoroughly in a couple days and get it merged and tagged.

I added a test in there as well to confirm it's working so I'm pretty sure everything's fine.

PR #144, opened 2021-11-29
fix(aliases): use providers to create aliases

QWp6t:
Attempting to inject an aliased class that hasn't yet been bound causes things to blow up.

I encountered this in a Sage project where SageServiceProvider was building app('sage'), which required an instance of `\Roots\Acorn\Filesystem\Filesystem`, but the filesystem had not yet been bound via the FilesystemServiceProvider. This resulted in the application container injecting \Illuminate\Filesystem\Filesystem, which is not an instance of \Roots\Acorn\Filesystem\Filesystem.

x-ref #142
x-ref #143


QWp6t:
I'm also removing Roots\Acorn\Config.

It would have actually suffered from the same injection issue mentioned in my previous post, so I would either have to fix that by modifying the config bootstrap to bind app('config') to an instance of the acorn Config class, or I can remove the alias entirely. I chose the latter, which left us with a class that was completely orphaned.

This class had no methods, and didn't seem to serve a purpose other than to have a Roots namespace. If there's a need for it down the road, we can re-add it.

PR #150, opened 2021-12-12
fix(console): use upstream `Kernel::load()` method

### LINKED ISSUE #149: Bug: Roots\Acorn\Console\Kernel does not autoload commands from App\Console\Commands
### 

#### What's wrong?

`Roots\Acorn\Console\Kernel` does not autoload commands from `App\Console\Commands`.

#### What have you tried?

Adding a command to `App\Console\Commands` and then watching it be ignored.

#### What insights have you gained?

The kernel only loads commands that subclass `Illuminate\Console\Commands\Command`, but that class doesn't exist.

#### Possible solutions

This line should actually be `use Illuminate\Console\Command;`: https://github.com/roots/acorn/blob/99062b1ade623786a73ce27427be4db79cfaa612/src/Roots/Acorn/Console/Kernel.php#L7

#### Temporary workarounds

🤷‍♀️

### Steps To Reproduce

1. Add a command to `App\Console\Commands`
1. Run `wp acorn`
1. See that it is missing

### Expected Behavior

It should appear in the command list.

### Actual Behavior

It does not.

### Relevant Log Output

_No response_

### Versions

99062b1ade623786a73ce27427be4db79cfaa612

stefanfisk:


QWp6t:
Thanks for the PR!

Turns out that method isn't needed at all and we can just let the parent class's method be called instead.

PR #173, opened 2022-01-31
✨ add acorn:dump-bootloader command

QWp6t:
This command will dump a bootloader file containing the resolved application paths.

This is useful to avoid resolving those paths on each request.

This command should be interpreted as a caching step, and the resulting file probably shouldn't be committed to your repo (consider adding it to your .gitignore).

Here's an example of the generated file:

image


QWp6t:
Gonna put this on hold for now.

Because it's trying to load from mu-plugins, it also means that it could be loading prior to autoloaders being registered inside of plugins and themes, making class resolution impossible causing providers to be skipped and exceptions to be thrown and other unexpected behavior.

I'll revisit this when I have more bandwidth.

QWp6t:
Closing this because its usage could cause bugs and confusion in some environments.

We include something similar by default in [Radicle](https://roots.io/radicle/).

PR #178, opened 2022-02-01
🧑‍💻 improve bud hmr compatibility

kellymears:
This fixes HMR in roots/bud on subpages. See: https://github.com/roots/bud/issues/1070. 

`webpack-hot-middleware` must use webpack's `output.publicPath` internally. which.. makes sense. But, it's not ideal here.

In the long term I might do something like fork whm or implement my own HMR solution.  It makes me super uncomfortable to not have control over the public path. So, if you want to think about it as a temporary workaround.. I just want roots/sage users to have a nice experience.

### DIFF
--- a/src/Roots/Acorn/Assets/Bundle.php
+++ b/src/Roots/Acorn/Assets/Bundle.php
@@ -121,10 +121,15 @@ class Bundle implements BundleContract
 
     protected function isBudRequest()
     {
-        if (!$path = realpath("{$this->path}/hmr.json")) return false;
-        if (!$header = request()->header('x-bud-dev-origin')) return false;
+        if (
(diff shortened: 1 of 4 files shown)


kellymears:
It makes sense that this test would fail. There is no way for PHP to know these assets are proxied and will be served by a node server.

src/Roots/Acorn/Assets/Bundle.php
```diff
@@ -125,7 +125,7 @@ protected function getUrl($path)
             return $path;
         }
 
-        return "{$this->uri}/${path}";
```
kellymears: Tell me you type template literals without telling me you type template literals.

PR #193, opened 2022-02-14
📦️ update Laravel v9.0.2

QWp6t:


LeoColomb:
Looks great! 😊
Naive question: wouldn't this be the opportunity to switch the modern exception page https://github.com/spatie/ignition (or even https://github.com/spatie/laravel-ignition)?

PR #211, opened 2022-03-19
🤝 add header for rest api requests when using bud

QWp6t:


QWp6t:
It looks like this is no longer required. We can revisit it another time if necessary.

PR #224, opened 2022-04-30
fix: Fixed Sage 10 view resolving as a child theme

LiamMartens:
When using a Sage theme as a child theme of a non Acorn theme the templates aren't being picked up correctly because they are being parsed as blade templates
This commit fixes that by treating non-blade templates "as-is"

Secondly - in this instance - when used as a child template we should probably use the stylesheet directory method otherwise the index.php file of the parent theme is included.

### DIFF
--- a/src/Roots/Acorn/Sage/Concerns/FiltersTemplates.php
+++ b/src/Roots/Acorn/Sage/Concerns/FiltersTemplates.php
@@ -27,6 +27,13 @@ trait FiltersTemplates
      */
     public function filterTemplateInclude($file)
     {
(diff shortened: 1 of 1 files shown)


QWp6t:
Non-blade temples should still get passed to Blade.

Our usage of index.php has changed, and it currently only acts as an entrypoint for loading Blade. All views are now located within the sage views/ directory, including the layout.

Our default view config also includes both parent and child themes.

https://github.com/roots/acorn/blob/main/config/view.php#L17-L18

If you see any other areas where we could better support child themes, please feel free to let us know.

PR #230, opened 2022-06-26
✨ add support for mjs enqueues

QWp6t:
I've only done some rough testing on this one, but I think it should be good to go.

### DIFF
--- a/src/Roots/Acorn/Sage/Concerns/FiltersEnqueues.php
+++ b/src/Roots/Acorn/Sage/Concerns/FiltersEnqueues.php
@@ -1,0 +1,33 @@
+<?php
+
+namespace Roots\Acorn\Sage\Concerns;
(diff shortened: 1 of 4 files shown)


src/Roots/Acorn/Sage/Concerns/FiltersEnqueues.php
```diff
@@ -0,0 +1,33 @@
...
+            return $tag;
+        }
+
+        // remove current type if it's defined
+        $tag = preg_replace('/type=["\'][^"\']+["\']/', '', $tag);
+
+        // now add module type
+        return str_replace('<script ', '<script type="module" ', $tag);
+    }
+}
```
kellymears, line 33: this is nice

kellymears (approved):

PR #236, opened 2022-07-05
fix(package-discovery): Check for packages in webroot (Bedrock)

Log1x:
Potential fix for https://discourse.roots.io/t/bedrock-required-composer-packages-not-registered-with-acorns-packagemanifest/23495

Tested and working on my end – not sure if there's a more sane approach.

### DIFF
--- a/src/Roots/Acorn/Application.php
+++ b/src/Roots/Acorn/Application.php
@@ -122,7 +122,6 @@ class Application extends FoundationApplication
                 throw new Exception("The {$path_type} path type is not supported.");
             }
 
-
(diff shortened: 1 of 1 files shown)


QWp6t (approved):

oxyc:
Another way to get the data would be

```php
$composer_paths = ...
    ->merge(
        collect(\Composer\InstalledVersions::getAllRawData())->pluck('install_path'),
    )
    ->...
```

But if doing that might as well use getAllRawData() to pick up the `['extra']['acorn']` configs as well.

I have nothing against assuming a bedrock like structure though :)

PR #239, opened 2022-07-08
🐛 Fix javascript enqueue handle

### LINKED ISSUE #238: Bug: Javascript not executing in production build since 2.1.0
### Terms

#### What's wrong?
In a Sage 10 + roots/acorn 2.1.0 install, after running yarn build, the scripts are enqueued but not executed.

#### What have you tried?
Downgrading bud to 5.x, no effect.

#### Temporary workarounds
Downgrading roots/acorn to v2.0.6.

### Steps To Reproduce

1. Clone roots/sage^10
2. Require roots/acorn@v2.1.0
3. Add some script to sage/resources/scripts/app.js
4. yarn build
5. View in browser

### Expected Behavior

Scripts (both app.js & editor.js) should run

### Actual Behavior

Scripts (both app.js & editor.js)  don't run

### Relevant Log Output

```shell
There are no console errors, no PHP errors, no acorn logs.
```

### Versions

2.1.0

Log1x:
- fix(bundle): Fix javascript enqueue handle (Fixes #238)
- chore(bundle): Add missing docblocks

### DIFF
--- a/src/Roots/Acorn/Assets/Bundle.php
+++ b/src/Roots/Acorn/Assets/Bundle.php
@@ -128,7 +128,7 @@ class Bundle implements BundleContract
     /**
      * Get the bundle URL.
      *
-     * @param  mixed $path
+     * @param  string $path
      * @return string
      */
     protected function getUrl($path)
--- a/src/Roots/Acorn/Assets/Bundle.php
+++ b/src/Roots/Acorn/Assets/Bundle.php
@@ -129,7 +129,7 @@ class Bundle implements BundleContract
      * Get the bundle URL.
      *
      * @param  mixed $path
-     * @return mixed
+     * @return string
      */
     protected function getUrl($path)
     {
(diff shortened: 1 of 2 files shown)


QWp6t (changes requested):
The problem with using `+` instead of `array_merge()` when it comes to indexed arrays is that it will clobber matching indexes on the 2nd array. So we'll have to figure out a more workable solution, I think, unless entrypoints.json will NEVER have js and mjs keys at the same time. (cc @kellymears)

image

image

kellymears:
@QWp6t If a user enables esm it likely will not emit `js` with defaults.

But there may be very good reasons to use mjs and js together, I'm really not sure. I think it's good to support mjs/js both, if possible. It would be pretty easy for users to make a couple small mods and end up in a situation where both keys are present.

Log1x:
Adding an `array_merge()` to `inline()` in `Enqueueable` seems to be the easiest solution. This will keep the array keys the same during enqueueing and inlining.

Screenshot

QWp6t (approved):
LGTM 🚀

PR #244, opened 2022-07-24
Update 1.x to Laravel 8 and minimum PHP to 7.3

joshuafredrickson:
This bumps the 1.x branch to Laravel 8 from 7, also increasing the minimum PHP version to 7.3.

This specifically addresses compatibility with newer versions of `log1x/acf-composer` as it uses `Str::remove`, which was introduced in Laravel 8.41.0.

Ref: https://github.com/Log1x/acf-composer/blob/33e2b7529fbd7190f024dd3d66a6e9a2b95e6f5b/src/AcfComposer.php#L91
Ref: https://github.com/laravel/framework/releases/tag/v8.41.0


QWp6t (changes requested):
Thanks for the PR!

This would actually be a breaking change in Acorn 1.x so a few changes would need to be made in order for it to be merged.

The PHP version constraint would have to remain as ^7.2.5|^8.0, and the illuminate/* packages would have to be ^7.0|^8.0.

But I also don't currently have the time to review and test that change to the extent that it ought to be reviewed and tested. If anyone else wants to volunteer to do that, let me know and we can see about getting this merged with those changes.

My personal opinion is that users should upgrade to Acorn v2 or wait for Acorn v3, though.

joshuafredrickson:
I've been thinking on this for a minute, and you're right. Probably the best path forward is to take a serious look into updating the v1 sites at some point. I don't have the bandwidth to test PHP 7.2 or 7.3 in this context either. 

@Log1x addressed this issue in `acf-composer`, which was my main reason for submitting this PR. 👍 Not to say that there aren't other compatibility issues that will arise in other situations. 

I have about 4-5 sites that are currently on v1 that I can use to test migration paths moving forward. In the interim, I'll post on Discourse to see if anyone has experience migrating from v1 to v2.

PR #246, opened 2022-07-28
fix: registerPackageManifest open_basedir restriction

### LINKED ISSUE #245: Bug: open_basedir restriction as a plugin without bedrock
### Terms

#### What's wrong?
open_basedir restriction in effect occurred after installing acorn as a plugin without bedrock when open_basedir restriction is present.

#### What have you tried?
- Have same issue when reproduced in my other server
- Tried to install it as theme dependency still getting same issue

#### What insights have you gained?
During acorn initialization, the `registerPackageManifest:270` method in `\src\Roots\Acorn\Application.php` attempted to access outside working directory which is outside of allowed path in open_basedir
https://github.com/roots/acorn/blob/32361ac69cbc7c87a34755c58fe79ec3f399ed8f/src/Roots/Acorn/Application.php#L269-L272

#### Possible solutions
Open to suggestions, however, this is my thought regarding the issue
1. Checking if a path is within open_basedir
2. Suppress the isFile() warning

#### Temporary workarounds
Currently, I'm doing the second solution as I did in [my fork](https://github.com/iniznet/acorn/blob/c70f1de799d3551f213a36a3363845b15bd6dd7a/src/Roots/Acorn/Application.php#L270-L271). I'm not sure if the suppression is the right way to do.
```
->filter(function ($path) use ($files) {
    return @$files->isFile("{$path}/vendor/composer/installed.json")
        && @$files->isFile("{$path}/composer.json");
})
```

### Steps To Reproduce

1. Setup server with open_basedir present with value set something like `/www/wwwroot/domain_working_dir/:/tmp/`
2. Install WordPress & Sage 10
3. Install Acorn as a plugin or theme dependency
4. Check and confirm if any page is accessible

### Expected Behavior

Able to load any page fine when open_basedir is present

### Actual Behavior

Got an open_basedir restriction warning after installing acorn as a plugin without bedrock when accessing any pages.

### Relevant Log Output

```shell
is_file(): open_basedir restriction in effect. File(/www/wwwroot/vendor/composer/installed.json) is not within the allowed path(s): (/www/wwwroot/domain_working_dir/:/tmp/)
```

### Versions

Acorn 2.1.2. Sage 10, Bud 6.3.3, WordPress 6.0.1, PHP 8.0, Ubuntu 20.4

iniznet:
This PR fix #245 by suppressing the isFile() method

### DIFF
--- a/src/Roots/Acorn/Application.php
+++ b/src/Roots/Acorn/Application.php
@@ -267,8 +267,8 @@ class Application extends FoundationApplication
                 })
                 ->unique()
                 ->filter(function ($path) use ($files) {
-                    return $files->isFile("{$path}/vendor/composer/installed.json")
-                        && $files->isFile("{$path}/composer.json");
+                    return @$files->isFile("{$path}/vendor/composer/installed.json")
+                        && @$files->isFile("{$path}/composer.json");
                 })
                 ->all();
 


QWp6t:
Thanks for the PR! 🙏

PR #251, opened 2022-10-17
⬆️ Update to Laravel v9.47.0

retlehs:
Laravel v9.47.0


davidwebca:
I just saw the new Blade's @fragment declarations in 9.39 and I'm already thinking about tightly integrating this to a set of Ajax actions that would allow to refresh only parts of a view without messing the main query or declaring custom ones. 👀

PR #268, opened 2023-02-07
Add relativePath to Asset contract

joshuafredrickson:


QWp6t:
This is technically a breaking change so gonna put this off until the next major version of Acorn.

PR #277, opened 2023-02-22
✨ Database support

### LINKED ISSUE #119: Illuminate/database support?
Hi guys,

I'm looking into Wordpress MVC frameworks and am really liking the Bedrock/Sage/Acorn combo. I noticed illuminate/database was added in March 2019 but removed in Nov 2019. Just wondering what the reason for this was and if there are any issues using Eloquent with Acorn?

Also, does the Roots stack have any routing and middleware options? Or was that intentionally excluded in favour of doing it the Wordpress way?

Thanks :)

retlehs:
This PR adds `illuminate/database` as a dependency and enables migration commands:

```shell
$ wp acorn migrate

$ wp acorn db:seed
$ wp acorn db:wipe 

$ wp acorn make:migration

$ wp acorn migrate:fresh
$ wp acorn migrate:install
$ wp acorn migrate:refresh
$ wp acorn migrate:reset
$ wp acorn migrate:rollback
$ wp acorn migrate:status
```

Seed command requires https://github.com/laravel/laravel/blob/9.x/database/seeders/DatabaseSeeder.php (and [loading it](https://github.com/laravel/laravel/blob/5b60b604c4167b30286a6e033723a6b274e8452d/composer.json#L27))

* * *

```shell
$ wp acorn make:migration create_test_table

   INFO  Migration [database/migrations/2023_02_22_050437_create_test_table.php] created successfully.

$ wp acorn migrate:status

  Migration name ............................................................. Batch / Status
  2023_02_22_050437_create_test_table ............................................... Pending

$ wp acorn migrate

   INFO  Running migrations.

  2023_02_22_050437_create_test_table .............................................. 9ms DONE
```

Closes #119


config/database.php
```diff
@@ -51,7 +51,7 @@
             'password'       => DB_PASSWORD,
             'unix_socket'    => env('DB_SOCKET', ''),
             'charset'        => DB_CHARSET,
-            'collation'      => DB_COLLATE,
+            'collation'      => $GLOBALS['wpdb']->collate,
             'prefix'         => $table_prefix ?? $GLOBALS['wpdb']->prefix,
```
QWp6t, line 55: `$table_prefix` will always be `null` here. We could either use `$GLOBAL['table_prefix']` or drop it entirely. I'm leaning toward the latter.

```suggestion
            'prefix'         => $GLOBALS['wpdb']->prefix,
```

QWp6t:
This introduces breaking changes in the console.

Someone can't update Acorn without running into issues in their console that are pretty difficult to recover. Since the console kernel crashes before fully booting, it's not possible run commands like `acorn optimize:clear` or `acorn clear-compiled`

image

The way to fix this is to conditionally load the command(s) added here. I will look into a solution that can be reused so we can more easily add support for other Laravel components without running into this problem.

config/database.php
```diff
@@ -51,8 +51,8 @@
             'password'       => DB_PASSWORD,
             'unix_socket'    => env('DB_SOCKET', ''),
             'charset'        => DB_CHARSET,
-            'collation'      => DB_COLLATE,
-            'prefix'         => $table_prefix ?? $GLOBALS['wpdb']->prefix,
+            'collation'      => $GLOBALS['wpdb']->collate,
+            'prefix'         => $GLOBALS['wpdb']->prefix,
```
retlehs, line 55: @QWp6t What are your thoughts on setting the `prefix` to just `''` instead of using the typical `wp_` prefix? It seems like it would make sense for things coming from Acorn to not use that prefix

For example, the default Laravel users migration example conflicts with the `wp_users` table

QWp6t: I think having the WordPress prefix makes sense here, actually.

For non-WordPress database stuff, we should be encouraging the use of the [`mysql`](https://github.com/roots/acorn/blob/main/config/database.php#L64-L82) connection instead of the [`wordpress`](https://github.com/roots/acorn/blob/main/config/database.php#L46-L62) connection. The idea with the `wordpress` connection would be to provide some type of ORM for WordPress, such as through the use of Corcel.

I'm open to any pushback on that idea, though, if you think I'm being a dumdum. :+1:

retlehs: @QWp6t That makes sense :+1:

chrillep:
oh sry btw will this enable factories and models?
or is that a future PR? =)

chrillep:
even better https://github.com/kitloong/laravel-migrations-generator!

QWp6t (approved):
LGTM! :shipit:

chrillep:
great news! good work @retlehs @QWp6t ❤️

chrillep:
i noticed [corcel](https://github.com/corcel/corcel) already had models for WP https://github.com/corcel/corcel/tree/7.0/src/Model

PR #278, opened 2023-02-22
📝 Supported components table

retlehs:


QWp6t:
This is pretty easy to miss. 

image

I think we should consider highlighting a few of the bigger ones, and then have a details view that is a more exhaustive list. What do you think? If that sounds good to you, I can try to whip up some changes to this PR on Friday.

retlehs:
@QWp6t Sounds good to me :+1:

PR #286, opened 2023-05-04
Added explicit property 'protected Filesystem ' to remove deprecated …

sammyhayman:
…notice in Sage class.

This should fix the "Creation of dynamic property Roots\Acorn\Sage\Sage::$files is deprecated" notice.

### DIFF
--- a/src/Roots/Acorn/Sage/Sage.php
+++ b/src/Roots/Acorn/Sage/Sage.php
@@ -43,7 +43,14 @@ class Sage
      *
      * @var ViewFactory
      */
-    protected $view;
+    protected          $view;
+
+    /**
+     * The Filesystem instance.
+     *
+     * @var Filesystem
+     */
+    private Filesystem $files;
 
     /**
      * Creates a new Sage instance.


src/Roots/Acorn/Sage/Sage.php
```diff
@@ -43,7 +43,14 @@ class Sage
...
      */
-    protected $view;
+    protected          $view;
+
+    /**
+     * The Filesystem instance.
+     *
+     * @var Filesystem
+     */
+    private Filesystem $files;
```
QWp6t, line 53: I try to avoid making things `private` unless there's compelling reason to do so.

```suggestion
    protected Filesystem $files;
```

QWp6t (changes requested):
Thank you for the PR! 🙏🏼 

Just a small change that I'm requesting and we'll be good to merge. 🙂

QWp6t (approved):
Awesome, this looks great! Thanks again! 🙏

src/Roots/Acorn/Sage/Sage.php
```diff
@@ -43,7 +43,14 @@ class Sage
      *
      * @var ViewFactory
      */
-    protected $view;
+    protected          $view;
```
QWp6t, line 46:
```suggestion
    protected $view;
```

PR #293, opened 2023-06-15
Hybrid fse

r-chrzan:
PR overwrites templates in site editor with those with blade php if it is in the theme


strarsis:
@r-chrzan: This appears to cover the same issue as https://github.com/roots/acorn/pull/141.

PR #295, opened 2023-06-20
#270 Removing a provider breaks Acorn until removing the packages cache

### LINKED ISSUE #270: Removing a provider breaks Acorn until removing the packages cache
Adding and then removing a service provider breaks Acorn until the packages cache is removed:

1. `composer require spatie/laravel-ignition`
2. `wp acorn package:discover`
3. `composer remove spatie/laravel-ignition`

Error:

```
In Application.php line 354:

  Skipping provider [Spatie\LaravelIgnition\IgnitionServiceProvider] because it does not exist.
```

Workaround:

1. `rm cache/acorn/framework/cache/packages.php`
2. `wp acorn optimize`

chrillep:
fixes #270

* tested on radicle

* needs more testing in other contexts

did some digging and laravels default method is not invoking a new Process of the command via cli. Its removing the cache files in the same instance.

https://github.com/laravel/framework/blob/e217dca49e20cd2840970f8d307c8392012bc5be/src/Illuminate/Foundation/ComposerScripts.php#L48-L69

While acorn is initiating new Process of wp acorn cli command. It fails returning exit code 1. Since the Provider may have been removed when executing. 

https://github.com/roots/acorn/blob/b020038947cd429d6020fc0bbc24568ffd9677d1/src/Roots/Acorn/ComposerScripts.php#L21

https://github.com/roots/acorn/blob/b020038947cd429d6020fc0bbc24568ffd9677d1/src/Roots/Acorn/Console/Console.php#L9-L17

https://github.com/roots/acorn/blob/b020038947cd429d6020fc0bbc24568ffd9677d1/src/Roots/Acorn/Console/Console.php#L99-L110

Possible solution. 

Reuse laravels method and make sure 

getCachedConfigPath() resolves to
`{projectpath}/storage/framework/cache/config.php`

getCachedServicesPath() resolves to
`{projectpath}/storage/framework/cache/services.php`

getCachedPackagesPath() resolves to
`{projectPath}/storage/framework/cache/packages.php`

instead of
`{projectpath}/bootstrap/cache/config.php`
`{projectpath}/bootstrap/cache/services.php`
`{projectPath}/bootstrap/cache/packages.php`


src/Roots/Acorn/ComposerScripts.php
```diff
@@ -15,9 +13,18 @@ class ComposerScripts extends FoundationComposerScripts
      */
     protected static function clearCompiled()
     {
-        $console = new Console(new Filesystem(), getcwd());
+        $laravel = new \Roots\Acorn\Application(getcwd(), ['bootstrap' => getcwd() .'/storage/framework']);
```
QWp6t, line 16: This path should not be hard-coded. Users can change the path by setting `ACORN_BOOTSTRAP_PATH` or calling `Application::useBootstrapPath()`. So we shouldn't assume any specific value or pattern.

You can try using the trait [`GetsFreshApplication::getFreshApplication()`](https://github.com/roots/acorn/blob/main/src/Roots/Acorn/Console/Concerns/GetsFreshApplication.php) if you want. Seems like that might be what you're going for here.

QWp6t (changes requested):
Thanks for the PR! 🙏

I haven't tested this, but I noticed an issue with your approach. Let me know if you want me to jump in and help get this over the finish line or if you prefer to tackle it.

chrillep:
> Thanks for the PR! 🙏
> 
> 
> 
> I haven't tested this, but I noticed an issue with your approach. Let me know if you want me to jump in and help get this over the finish line or if you prefer to tackle it.

Yeah that would be great! You jump in and get it over the finish line. 😀

I just wanted to get the ball rolling with what I found.

src/Roots/Acorn/ComposerScripts.php
chrillep: ooh i didnt know

retlehs:
This PR doesn't work for me :eyes: Steps tested on a regular Bedrock install:

1. `composer require roots/acorn`
2. `composer require roots/acorn-prettify`
3. `wp acorn optimize:clear`
4. Pull in the changes from this PR
5. `composer remove roots/acorn-prettify`
6. `wp acorn`

```
Skipping provider [Roots\AcornPrettify\AcornPrettifyServiceProvider] because it does not exist.
```

PR #296, opened 2023-06-24
🧑‍💻 Add .devcontainer

QWp6t:
It's not perfect. I'd like to iterate on this in the future. But it's working, and I'm finding it useful so far.

It runs nginx in a separate container, which was a goal of mine.

It provides other services as well in case we have a desire to test them.

It is compatible with both Bedrock and Radicle.
Copy .devcontainer/config/app/.env.example to .devcontainer/.env and append `REPOSITORY_URL='git@github.com:roots/radicle.git'` to test with Radicle.

Few ideas for the future
- Figure out how to get SSH forwarding to work. Right now I just set ~/.ssh as readonly volume to ~/.ssh in the container lol
- Clean up the scripts. Maybe use something similar to what Trellis does with its deployment-hooks.
- Possibly look into integrating directly with Trellis.
- Make it easier for developers to extend or modify the devcontainer without resulting in a dirty repo.
- Add SSL support
- Run node in a separate container

Credits

This was my first time messing with devcontainers. I borrowed from and learned from other examples.

- https://github.com/laravel/sail/
- https://github.com/Cyber-Duck/php-fpm-laravel/tree/8.2
- https://github.com/acadea/laravel-sail-nginx-php-fpm/
- https://github.com/wordpress/wordpress-develop/tree/trunk/.devcontainer

### DIFF
--- a/.devcontainer/config/web/default.conf
+++ b/.devcontainer/config/web/default.conf
@@ -1,0 +1,50 @@
+server {
+  listen 80 default;
+  listen [::]:80;
+
+  # listen 443 ssl;
+  # listen [::]:443 ssl ipv6only=on;
+  # ssl_certificate /etc/nginx/ssl/default.crt;
+  # ssl_certificate_key /etc/nginx/ssl/default.key;
+
+  add_header X-Frame-Options "SAMEORIGIN";
+  add_header X-Content-Type-Options "nosniff";
+  charset utf-8;
+
+  server_name web.local web;
+  root /roots/app/public;
+  index index.php index.html;
+
+  error_log  /dev/stderr info;
+  access_log /dev/stdout;
+
+  client_max_body_size 100M;
+
+  location / {
+    try_files $uri $uri/ /index.php$is_args$args;
+  }
+
+  location ~ \.php$ {
+    try_files $uri /index.php =404;
+    fastcgi_pass php-upstream:9000;
+    fastcgi_index index.php;
+    fastcgi_buffers 16 16k;
+    fastcgi_buffer_size 32k;
+    fastcgi_param SCRIPT_FILENAME $document_root$fastcgi_script_name;
+    fastcgi_read_timeout 600;
+    include fastcgi_params;
+  }
+
+  if (!-e $request_filename) {
+    rewrite ^.*$ /index.php last;
+  }
+
+  location = /favicon.ico {
+    access_log off;
+    log_not_found off;
+  }
+  location = /robots.txt  {
+    access_log off;
+    log_not_found off;
+  }
+}
--- a/.devcontainer/docker-compose.yml
+++ b/.devcontainer/docker-compose.yml
@@ -1,0 +1,144 @@
+version: '3.9'
+
+services:
+  acorn.test:
+    build:
+      context: .
+      dockerfile: Dockerfile
+    image: roots/dev-8.2
+    extra_hosts:
+      - 'host.docker.internal:host-gateway'
+    environment:
+      PHP_IDE_CONFIG: '${PHP_IDE_CONFIG:-serverName=acorn}'
+      XDEBUG_MODE: '${XDEBUG_MODE:-develop,debug}'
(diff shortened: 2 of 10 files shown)


.devcontainer/config/web/default.conf
```diff
@@ -0,0 +1,50 @@
+server {
+  listen 80 default;
+  listen [::]:80;
+
+  # listen 443 ssl;
+  # listen [::]:443 ssl ipv6only=on;
+  # ssl_certificate /etc/nginx/ssl/default.crt;
+  # ssl_certificate_key /etc/nginx/ssl/default.key;
```
QWp6t, line 8: This is untested.

.devcontainer/docker-compose.yml
```diff
@@ -0,0 +1,144 @@
...
+    image: roots/dev-8.2
+    extra_hosts:
+      - 'host.docker.internal:host-gateway'
+    environment:
+      PHP_IDE_CONFIG: '${PHP_IDE_CONFIG:-serverName=acorn}'
+      XDEBUG_MODE: '${XDEBUG_MODE:-develop,debug}'
+      XDEBUG_CONFIG: '${XDEBUG_CONFIG:-client_port=9003 client_host=host.docker.internal discover_client_host=true}'
+      REPOSITORY_URL: '${REPOSITORY_URL:-https://github.com/roots/bedrock.git}'
+    volumes:
+      - ~/.ssh:/home/vscode/.ssh:ro
```
QWp6t, line 17: I tried using SSH agent forwarding, but it didn't work for me. I tried running an agent in Windows. I tried while running an agent in WSL. I tried with an agent running in both Windows and WSL. In all scenarios, the agent was confirmed running but I did not have access to GitHub inside the container.

So this line is here as a workaround, with an understanding that this is a rather limited approach.

.devcontainer/install.sh
```diff
@@ -0,0 +1,102 @@
...
+cd /roots/app
+
+# if composer.json already exists, exit early
+if [ -f "composer.json" ]; then
+  exit 0
+fi
+
+# Clone the repository
+git clone --depth=1 ${REPOSITORY_URL} . \
+  && rm -rf .git
```
QWp6t, line 36: This could _maybe_ be an issue if `REPOSITORY_URL` points to a path on the local file system. I'm not sure; I haven't tested that.

.devcontainer/install.sh
```diff
@@ -0,0 +1,102 @@
...
+elif cat "${WORKSPACE_FOLDER}/composer.json" | jq '.type' | grep -q wordpress-plugin; then
+  ln -fs "${WORKSPACE_FOLDER}" "/roots/app/public/app/plugins/$(basename ${WORKSPACE_FOLDER})"
+elif cat "${WORKSPACE_FOLDER}/composer.json" | jq '.type' | grep -q wordpress-muplugin; then
+  ln -fs "${WORKSPACE_FOLDER}" "/roots/app/public/app/mu-plugins/$(basename ${WORKSPACE_FOLDER})"
+else
+  cat /roots/app/composer.json | jq ".repositories += [{ type: \"path\", url: \"${WORKSPACE_FOLDER}\" }]" > /roots/app/composer.tmp \
+  && rm /roots/app/composer.json \
+  && mv /roots/app/composer.tmp /roots/app/composer.json \
+  && composer require -d /roots/app $(cat "${WORKSPACE_FOLDER}/composer.json" | jq '.name' | tr -d '"')
+fi
```
QWp6t, line 81: This can help make the devcontainer more portable/reusable outside of Acorn.

.devcontainer/setup.sh
```diff
@@ -0,0 +1,60 @@
+#!/bin/sh
+
+set -eux
+
+# Install volta
+curl https://get.volta.sh | bash -s -- --version latest
+export VOLTA_HOME="${HOME}/.volta"
+export PATH="${VOLTA_HOME}/bin:${PATH}"
+volta install node
+volta install yarn
```
QWp6t, line 10: The idea here is that we can test theme frontend stuff using multiple versions of node. Again, this is intended to be reused at some point.

PR #301, opened 2023-07-27
Exit parse_request to short circuit wordpress during custom routes.

eavonius:
This is the shortest pull request of all time ;). But it seems to work as described by the people who found it on discourse.

Fixes the issue described here:

https://discourse.roots.io/t/custom-route-includes-the-homepage-view/25462/5


QWp6t:
Oh, wow. Thanks for this! 🙏

PR #282, opened 2023-03-28
PHP 8.2: ${var} string interpolation deprecated

firescript:
On a fresh install of PHP 8.2 and running Acorn inside my theme, i receive these notifications in the debug.log and /wp-admin dashboard alerts.

```
[28-Mar-2023 17:27:36 UTC] PHP Deprecated:  Using ${var} in strings is deprecated, use {$var} instead in /app/web/app/themes/firescript-cms/vendor/roots/acorn/src/Roots/Acorn/PackageManifest.php on line 67
[28-Mar-2023 17:27:36 UTC] PHP Deprecated:  Using ${var} in strings is deprecated, use {$var} instead in /app/web/app/themes/firescript-cms/vendor/roots/acorn/src/Roots/Acorn/PackageManifest.php on line 75
[28-Mar-2023 17:27:36 UTC] PHP Deprecated:  Using ${var} in strings is deprecated, use {$var} instead in /app/web/app/themes/firescript-cms/vendor/roots/acorn/src/Roots/Acorn/PackageManifest.php on line 114
```
This change fixes the notice.

https://php.watch/versions/8.2/$%7Bvar%7D-string-interpolation-deprecated#dollar-outside

PR #310, opened 2023-11-02
Removed `{}` to avoid deprecated warning

dmitrychuba:


Log1x:
From my understanding, `{$example}` isn't deprecated, `${example}` is. Are you actually seeing deprecation warnings?

dmitrychuba:
Yeah, I’ve noticed in the code it was like “{$example}” which appears to be correct. But still saw following warning: 
￼

Also my editor shows this Unnecessary curly braces message:
￼

Log1x:
Oh I see. This was fixed in https://github.com/roots/acorn/pull/282 but hasn't been pushed into a release yet. I'll see if we can get something out soon. 👍

PR #311, opened 2023-11-02
Create FoundationServiceProvider.php

tgeorgel:
Radicle (and so, Acorn) is missing the `FoundationServiceProvider`, causing multiple errors : 

- `\Illuminate\Http\Request` class is missing the `validate()` method
- The custom [FormRequests](https://laravel.com/docs/9.x/validation#form-request-validation) are not working, because of the missing call to the related service provider
- The dumper on Radicle is broken

By adding this Service provider to the `config/app.php` file, it fixes those errors.

```php
'providers' => [
    ...
    Roots\Acorn\Providers\FoundationServiceProvider::class,
],
```

Tested on Radicle stack.

### DIFF
--- a/src/Roots/Acorn/Providers/FoundationServiceProvider.php
+++ b/src/Roots/Acorn/Providers/FoundationServiceProvider.php
@@ -1,0 +1,25 @@
+<?php
+
+namespace Roots\Acorn\Providers;
+
+use Illuminate\Foundation\Providers\FormRequestServiceProvider;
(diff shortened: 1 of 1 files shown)


QWp6t:
FoundationServiceProvider is [already included](https://github.com/laravel/framework/blob/10.x/src/Illuminate/Support/DefaultProviders.php#L31) as part of the default providers.

I believe you created your PR before this change was added, which is why you might have previously experienced issues.

If you need to add any other providers to your config, you can do so in [config/app.php](https://github.com/roots/acorn/blob/main/config/app.php#L161-L168).

PR #313, opened 2023-11-10
'enqueueJs()' support to registering scripts with `async` and `defer`

ouun:
[WordPress 6.3 introduced support](https://make.wordpress.org/core/2023/07/14/registering-scripts-with-async-and-defer-attributes-in-wordpress-6-3/) for registering scripts with async and defer attributes. This PR updates `enqueueJs()` to support the new function signatures, with the prior `$in_footer` boolean parameter being overloaded to accept a new `$args` array parameter. 

```
    bundle('app')->enqueueJs([
        'strategy' => 'defer',
        'in_footer' => true,
    ]);
```

This is backwards compatible by core:

> For prior/existing usage of the wp_register_script() and wp_enqueue_script() functions making use of the $in_footer boolean param, backward compatibility is retained via logic that explicitly sets the scripts group to the applicable value for footer or header printing based on the boolean value passed to the new/overloaded $args parameter.


Log1x:
Can we update this so it defaults to a `bool` (specifically, `true`) so it doesn't break compatibility on earlier versions of WordPress?

PR #320, opened 2023-12-08
✨ Automatically expose public properties and methods on View Composers

Log1x:
This gives Acorn's View Composers the same behavior as View Components exposing all public properties/methods automatically to the targeted view(s).

An added bonus to this is being able to easily invoke public methods using the variable inside a view such as `$example($post->ID)`.

Most code is taken from [Illuminate's Component.php](https://github.com/illuminate/view/blob/master/Component.php#L220-L342) and moved into a Trait.

This only takes effect when `with()` and `override()` are empty thus maintaining backwards compatibility.

Could use more testing.

## Change log

- ✨ Automatically expose public properties and methods on View Composers
- 💥 Change the `$ignore` property to `$except` in the View Composer `Arrayable` trait

### DIFF
--- a/src/Roots/Acorn/View/Composers/Concerns/Extractable.php
+++ b/src/Roots/Acorn/View/Composers/Concerns/Extractable.php
@@ -37,15 +37,9 @@ trait Extractable
             $reflection = new ReflectionClass($this);
 
             static::$propertyCache[$class] = collect($reflection->getProperties(ReflectionProperty::IS_PUBLIC))
-                ->reject(function (ReflectionProperty $property) {
-                    return $property->isStatic();
-                })
-                ->reject(function (ReflectionProperty $property) {
-                    return $this->shouldIgnore($property->getName());
-                })
-                ->map(function (ReflectionProperty $property) {
-                    return $property->getName();
-                })->all();
+                ->reject(fn (ReflectionProperty $property) => $property->isStatic() || $this->shouldIgnore($property->getName()))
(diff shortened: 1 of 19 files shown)


src/Roots/Acorn/View/Composers/Concerns/Extractable.php
```diff
@@ -0,0 +1,101 @@
...
+
+namespace Roots\Acorn\View\Composers\Concerns;
+
+use Closure;
+use Illuminate\View\InvokableComponentVariable;
+use ReflectionClass;
+use ReflectionMethod;
+use ReflectionProperty;
+
+trait Extractable
```
QWp6t, line 11:
```suggestion
/**
 * Extract public properties from a composer to make available in a view. 
 *
 * @copyright Taylor Otwell
 * @link https://github.com/illuminate/view/blob/v10.35.0/Component.php#L220-L342
 */
trait Extractable
```

src/Roots/Acorn/View/Composers/Concerns/Extractable.php
```diff
@@ -0,0 +1,107 @@
...
+namespace Roots\Acorn\View\Composers\Concerns;
+
+use Closure;
+use Illuminate\View\InvokableComponentVariable;
+use ReflectionClass;
+use ReflectionMethod;
+use ReflectionProperty;
+
+/**
+ * Extract public properties from a composer to make available in a view.
```
QWp6t, line 12:
```suggestion
 * Extract public properties from a composer to make available in a view.
```

PR #330, opened 2024-01-23
Fix: Allow 4** HTTP responses to be rendered normally by router

### LINKED ISSUE #288: Bug: Routes incompatible with ValidationException
Acorn's route handling is incompatible with `ValidationException` because it throws a `422`

https://github.com/roots/acorn/blob/79eb9b02b59f7965d3cc227209dd9edf8e8058f4/src/Roots/Acorn/Bootloader.php#L230-L232

### Steps To Reproduce

1. Enable Acorn's routing
1. Add a route that will return a `422`:

```php
Route::get('/foo', function () {
    try {
        throw new \Illuminate\Validation\ValidationException(
            validator(data: [])
        );
    } catch (\Illuminate\Validation\ValidationException $exception) {
        return response()->json(data: [
            'message' => $exception->getMessage(),
            'errors' => $exception->errors(),
        ], status: $exception->status); // if you change status code to 200, it will work. Otherwise, it will be handled by WordPress
    }
});
```

h/t @thebezzo

oskar-koli:
Filtering out 4** status codes from the router responses seems unreasonable, it for example stops the returning of 401 Bad Request responses.

Fixes #288

### DIFF
--- a/src/Roots/Acorn/Bootloader.php
+++ b/src/Roots/Acorn/Bootloader.php
@@ -240,10 +240,6 @@ class Bootloader
             /** @var \Illuminate\Http\Response */
             $response = $kernel->handle($request);
 
-            if (! $response->isServerError() && $response->status() >= 400) {
-                return;
-            }
-
             $body = $response->send();
 
             $kernel->terminate($request, $body);


retlehs:
Is any existing 404 behavior affected by this change?

oskar-koli:
It shouldn't and based on my testing it doesn't.
If the path being accessed doesn't have a route defined in the router, then the code affected by this PR is never reached because the `bootHttp` function already bails out at [line 228](https://github.com/roots/acorn/blob/3.x/src/Roots/Acorn/Bootloader.php#L228). And so the route is after that handled normally by WP, resulting in the 404 page (assuming no normal WP route matches it).

retlehs (approved):

retlehs:
Thank you! Would you like to also submit a PR for this same change against the `main` branch as well?

PR #334, opened 2024-01-26
PSR-7 & PSR-15 support for WordPress Routes

broskees:
This PR add [PSR-7](https://www.php-fig.org/psr/psr-7/) & [PSR-15](https://www.php-fig.org/psr/psr-15/) support for WordPress Routes.

By using output buffering to capture WordPress's output and make it the body of a `Response`, we can force WordPress requests to actually emit a `Response` class. This allows PSR-15 (Middleware) and and PSR-7 (`Request`/`Response` classes) compatibility for WordPress itself. This works EVEN if WordPress uses output buffering as well, since output buffering can handle nested calls [(see here)](https://onlinephp.io?s=s7EvyCjg5eLlyk-KLy5JLCrR0LQGcVOTM_IVlDxSc3LyFcLzi3JSlKxxqkrMS1EoLkhMTlUoyc9XhChUyS8tKSgt8UstSy1yzc0sKUlNUbBVAGpPTy2JT85JTcxDNgJNHAA%2C&v=8.1.24).

> WordPress requests are already a Laravel Route, why is this necessary also?

For Middleware to be fully functional, they should also be able to make changes to the `Response` before it's sent to the client (See Laravel's documentation for an example of this: https://laravel.com/docs/10.x/middleware#before-after-middleware). This is impossible unless there is an actual `Response` class to manipulate.

Additionally, I'm improving on my last PR by using named routes instead of using a fake `JsonResponse` to identify WordPress requests like in my last PR. This lowers our maintenance burden and uses Laravel's native route matching to identify WordPress routes instead.


QWp6t:
@Log1x, @broskees 

Does my refactor make sense to y'all? If so, we can merge this. (I just didn't want to do a significant refactor without another pair of eyes peepin' it.)

Log1x:
> Does my refactor make sense to y'all? If so, we can merge this. (I just didn't want to do a significant refactor without another pair of eyes peepin' it.)

LGTM :shipit:

Issue #335, opened 2024-01-30
Latest changes to Bootloader::registerWordPressRoute() throws Exception


dsturm:
### Version

dev-main

### What did you expect to happen?

Site could be opened

### What actually happens?

The latest changes from #334 throw an exception for me - even on a clean `bedrock` /  `sage` project:

`Roots\Acorn\Bootloader::Roots\Acorn\{closure}(): Argument #1 ($response) must be of type Illuminate\Http\Response, Illuminate\Routing\ResponseFactory given, called in /sites/acorn-test/vendor/illuminate/support/helpers.php on line 307`

since [`response()`](https://github.com/broskees/acorn/blob/991bf7449f391b3afa2e5fa611fb84b52987e214/src/Roots/Acorn/Bootloader.php#L225) will return `Illuminate\Routing\ResponseFactory` rather than required `Illuminate\Http\Response` if arguments are omitted.

Further the [`header_remove()`](https://github.com/broskees/acorn/blob/991bf7449f391b3afa2e5fa611fb84b52987e214/src/Roots/Acorn/Bootloader.php#L227) fails because of already sent headers.

### Steps to reproduce

1. Install new `roots/bedrock` project
2. Require `roots/acorn:dev-main`
3. Install fresh `roots/sage` or `roots/sage:dev-main` into theme folder, install dependencies and build theme 
4. Visit front page

### Please confirm this isn't a support request.

Yes

PR #336, opened 2024-01-30
Fix: response() returning ResponseFactory instead of Response

dsturm:
This PR attempts to resolve issue #335, which involves an exception occurring in the new WordPress Routes support introduced in #334. 

Previously, when `response()` was called without any parameters, it would return `ResponseFactory::class` because of a [conditional check](https://github.com/roots/acorn/blob/9de4924933dc1e382576eff3b598287f102e232c/src/Illuminate/Foundation/helpers.php#L792). However, to obtain the required `Response::class`, an empty string now needs to be passed.

Furthermore, this pull request includes a check to prevent errors with `header_remove()` by verifying if headers have already been sent.


retlehs (approved):
Thank you! ❤️

PR #339, opened 2024-02-01
🐛 Bootloader: Bail early on `wp-login.php`/`wp-admin`, and some other `/wp-` requests

### LINKED ISSUE #337: Changes to WordPress routes break the wp-admin login
### Version

v4.0.0

Login into backend works as usual

### What actually happens?

After submitting the login credentials, I'll be "redirected" to the login form. On first attempt WordPress hints that cookies might be forbidden / deactivated.

### Steps to reproduce

- Upgrade to v4.0.0
- Try to login from `wp-login.php`

retlehs:
* [x] Fix wp-login.php
* [x] Fix wp-admin / admin-ajax.php requests
* [x] Fix other requests? wp-comments-post.php, wp-signup.php?

Fix #337

### DIFF
--- a/src/Roots/Acorn/Bootloader.php
+++ b/src/Roots/Acorn/Bootloader.php
@@ -261,12 +261,12 @@ class Bootloader
         ?\Illuminate\Routing\Route $route,
         array $config
     ) {
-        if (
-            strpos($request->getRequestUri(), '/wp-comments-post.php') !== false ||
-            strpos($request->getRequestUri(), '/wp-login.php') !== false ||
-            strpos($request->getRequestUri(), '/wp-signup.php') !== false ||
-            strpos($request->getRequestUri(), '/wp-admin/') !== false
-        ) {
+        if (Str::contains($request->getRequestUri(), [
+            '/wp-comments-post.php',
+            '/wp-login.php',
+            '/wp-signup.php',
+            '/wp-admin/',
+        ])) {
             return; // Let WordPress handle these requests
         }
 
--- a/src/Roots/Acorn/Bootloader.php
+++ b/src/Roots/Acorn/Bootloader.php
@@ -261,10 +261,14 @@ class Bootloader
         ?\Illuminate\Routing\Route $route,
         array $config
     ) {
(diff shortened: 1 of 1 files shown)


src/Roots/Acorn/Bootloader.php
```diff
@@ -261,6 +261,14 @@ protected function registerRequestHandler(
         ?\Illuminate\Routing\Route $route,
         array $config
     ) {
+        if (strpos($request->getRequestUri(), '/wp-login.php') !== false) {
+            add_filter('do_parse_request', function ($doParse, \WP $wp, $extraQueryVars) {
+                return false;
+            }, 100, 3);
```
QWp6t, line 267:
```suggestion
            add_filter('do_parse_request', '__return_false', 100, 3);
```

dsturm:
Thanks for starting a fix so quickly.

I also noticed that there were problems with REST queries in the block editor. But I'm not sure if it's just the cookie not being set.

retlehs:
Pushed an update that should fix those I think? If not, do you mind showing which requests are failing for you still?

dsturm:
@retlehs Worked for me! Thanks 💪

QWp6t (approved):
LGTM! :shipit:

dr5hn:
@QWp6t //cc @retlehs Doesn't work if we've hided login behind some custom path.

retlehs:
@dr5hn Does `get_admin_url()` return the correct URL on your setup?

dr5hn:
@retlehs Yes it return the hidden admin path url.

https://x***x.com/secret/

When i make this change, login works!

```
diff --git a/src/Roots/Acorn/Bootloader.php b/src/Roots/Acorn/Bootloader.php
index fc0e8c9..b57adfc 100644
--- a/src/Roots/Acorn/Bootloader.php
+++ b/src/Roots/Acorn/Bootloader.php
@@ -266,6 +266,7 @@ class Bootloader
             '/wp-login.php',
             '/wp-signup.php',
             '/wp-admin/',
+            '/secret/'
         ])) {
             return; // Let WordPress handle these requests
         }
```

PR #340, opened 2024-02-02
Fix `is_file()` deprecation warning

joshuafredrickson:
Using Sage on vanilla WP, you'll receive the following notice due to `Filesystem::closest` returning `null` because there's no `.env` to be found.

```
Deprecated: is_file(): Passing null to parameter #1 ($filename) of type string is deprecated in /vendor/roots/acorn/src/Roots/Acorn/Bootloader.php on line 376
```

### DIFF
--- a/src/Roots/Acorn/Bootloader.php
+++ b/src/Roots/Acorn/Bootloader.php
@@ -373,7 +373,7 @@ class Bootloader
      */
     protected function environmentPath(): string
     {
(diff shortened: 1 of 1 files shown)


retlehs (approved):
Thank you for the PR! This also happens on any environment that doesn't use an `.env` file even on Bedrock installs, like when using Lando

PR #343, opened 2024-02-05
Replace deprecated CORS config identifier class with HandleCors middleware

dsturm:
Since the `fruitcake/laravel-cors` is [deprecated since Laravel 9.x](https://github.com/laravel/laravel/pull/5825/files), we should replace the config identifier with the core one (`\Illuminate\Http\Middleware\HandleCors`).

### DIFF
--- a/src/Roots/Acorn/Providers/AcornServiceProvider.php
+++ b/src/Roots/Acorn/Providers/AcornServiceProvider.php
@@ -21,7 +21,7 @@ class AcornServiceProvider extends ServiceProvider
      * @var string[]
      */
     protected $providerConfigs = [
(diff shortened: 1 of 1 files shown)


src/Roots/Acorn/Providers/AcornServiceProvider.php
```diff
@@ -21,7 +21,7 @@ class AcornServiceProvider extends ServiceProvider
      * @var string[]
      */
     protected $providerConfigs = [
-        \Fruitcake\Cors\CorsServiceProvider::class => 'cors',
+        \Illuminate\Http\Middleware\HandleCors::class => 'cors',
```
QWp6t, line 24: This line should be removed entirely since it's not a service provider.

```suggestion
```

PR #344, opened 2024-02-05
Add `wp-cron.php` to exclusion list for request handler

dsturm:
This PR includes ﻿`wp-cron.php` request in the exclusion list for the request handler introduced in #339.

### DIFF
--- a/src/Roots/Acorn/Bootloader.php
+++ b/src/Roots/Acorn/Bootloader.php
@@ -263,6 +263,7 @@ class Bootloader
     ) {
         if (Str::contains($request->getRequestUri(), [
             '/wp-comments-post.php',
+            '/wp-cron.php',
             '/wp-login.php',
             '/wp-signup.php',
             '/wp-admin/',


iniznet:
Wouldn't it be better if we also allowed people to customize this through a filter hook, like:
```php
if (Str::contains(
    $request->getRequestUri(),
    apply_filters('acorn/router/ignore_request_paths', [
        '/wp-comments-post.php',
        '/wp-cron.php',
        '/wp-login.php',
        '/wp-signup.php',
        '/wp-admin/',
    ])
)) {
    return; // Let WordPress handle these requests
}
```

retlehs:
A filter is a good idea :+1: We also want to move away from the hardcoded paths and use `admin_url()`, `login_url()`, etc

@dsturm Were wp-cron.php requests failing for you without this change?

retlehs:
The paths added in #339 are for files that rely on the use of cookies, so `wp-cron.php` shouldn't be affected

@broskees mentioned on our Discord server that he might have a different fix to replace the code introduced in #339, since the response classes handle cookies too and the bootloader changes aren't accounting for that

Issue #342, opened 2024-02-05
www subdomain showing blank page


tombroucke:
### Version

4.0.1

### What did you expect to happen?

When visiting https://www.example.com, I should be redirected to https://example.com

After downgrading to v4.0.0-beta.1, the redirect works just fine.

### What actually happens?

There is no redirect, https://www.example.com shows a blank page.

### Steps to reproduce

I can't reproduce on my local machine (Valet + PHP 8.2). I'm experiencing this in production (Apache + PHP 8.2).
EDIT: The issue is reproducable on my local machine (LocalWP + PHP 8.2). New install of bedrock + sage.

### Please confirm this isn't a support request.

Yes

tombroucke:
As a temporary workaround, you can add the redirect in the .htaccess file

tombroucke:
So after doing a little digging, I found out that WordPress is doing a canonical redirect from whatever URL, to the URL that is set as site url (from `add_action( 'template_redirect', 'redirect_canonical' )`). 

Acorn's Bootloader's registerRequestHandler will take over this request, but it probably shouldn't. I suggest we add a guard clause:
```
```

PR #346, opened 2024-02-05
fix: Let WP handle requests that should trigger a canonical redirect

tombroucke:
The `registerRequestHandler()` method should let WordPress handle request that will result in a canonical redirect.

see https://github.com/roots/acorn/issues/342


retlehs (approved):
Thank you!

PR #350, opened 2024-02-08
Prevent with() and override() from being triggered twice on merge

folbert:
Store the outcome of with() and override() in variables instead of risking them being executing twice.


QWp6t:
Thanks for the PR @folbert! 🙏

PR #363, opened 2024-03-28
🔨 refactor(manifest): Throw custom error when bundle is not found in manifest

tombroucke:
I often get the general `Undefined array key "fonts"` exception when trying to enqueue a bundle before adding it to my `bud.config.js`. This PR replaces that general exception with a custom `BundleNotFoundException` with a more vebose message `Bundle [fonts] not found in manifest`,


QWp6t:
This looks great. Thank you! 🙏🏼

PR #367, opened 2024-03-29
🧑‍💻 Change default `editorStyles` implementation from URL to local theme path

Log1x:
ref: https://discourse.roots.io/t/absolute-domain-relative-path-for-asset-urls-in-css/23121/4


QWp6t (approved):
LGTM! :shipit:

PR #314, opened 2023-11-16
✨ Add support for (hybrid) FSE

dsturm:
While we already have a pull request (PR) for full-site editing (FSE) support by @strarsis (#141), this PR further enhances these modifications and now supports hybrid templates (Blade-PHP and FSE block templates, currently with priority given to the latter).

### Requirements theme

To enable FSE support, the theme (`sage`) **requires** some modifications:

1. Since we use `wp_is_block_theme()` to check for FSE themes when altering the theme hierarchy, `templates/index.html` needs to exist.
2. Currently `sage` [removes theme support for FSE](https://github.com/roots/sage/blob/be5281cdd178cb1faba9d8bba8185d6f6c1c4caf/app/setup.php#L52) - we need to remove / comment out this line. 
3. Because the generated CSS for blocks will be created in `wp_head()` action hook, the [current position](https://github.com/roots/sage/blob/be5281cdd178cb1faba9d8bba8185d6f6c1c4caf/index.php#L14) of the `view(...)->render()` method is too late and would result in empty styles for rendered blocks. Therefor, we need to [call it prior to `wp_head()`, store it in a temporary variable and echo it later](https://github.com/dsturm/sage/blob/5912c37dad81dcc2770b6f809f6b5ffa79ab4ac1/index.php#L6).

### Test this PR

Optionally create a new Bedrock project
```bash
composer create-project roots/bedrock
```

Use / require acorn fork and setup (minimal) prepared sage theme
```bash
# Require PR / fork
composer config repositories.acorn-fse vcs https://github.com/dsturm/acorn
composer require roots/acorn:dev-fse
# Clone new (sage) theme
cd web/app/themes
git clone -b fse https://github.com/dsturm/sage sage-fse
cd sage-fse
# Build
composer i -o && yarn && yarn build
# Activate theme
wp theme activate sage-fse
```

### Ideas

To use block patterns like `header` or `footer` in blade templates / views, I wrote [two directives](https://github.com/dsturm/sage/blob/5912c37dad81dcc2770b6f809f6b5ffa79ab4ac1/app/Providers/ThemeServiceProvider.php#L34) which are currently located in my [prepared `sage` fork](https://github.com/dsturm/sage), but should be located in `roots/acorn`:

[**@blocktemplate('template-part')**](https://github.com/dsturm/sage/blob/5912c37dad81dcc2770b6f809f6b5ffa79ab4ac1/app/Providers/ThemeServiceProvider.php#L39)

[...]

PR #368, opened 2024-03-30
🩹 Only prioritize block templates in block themes

Log1x:
Block templates should only be prioritized when there is a `index.html` (see: https://developer.wordpress.org/reference/classes/wp_theme/is_block_theme/)

Fixes potential issues like https://discourse.roots.io/t/acorn-v4-1-1-released/26905/2


Log1x:
Looking back at [the code](https://github.com/roots/acorn/pull/314/files#diff-7c4b19f127685bd366fb286df8b4f1568c58b17cbf5f7b56d63819ff611cfbc5R17-R47) on #314 in light of the issue reported above, things make a little more sense to me now.

I have reimplemented @dsturm's original approach. Apologies for the oversight.

PR #371, opened 2024-04-02
🩹 Fix duplicate provider registration

Log1x:
Noticed while working on #370 that providers are getting duplicated in the `provider.php` cache. This fixes it.


QWp6t (approved):
LGTM! :shipit:

PR #390, opened 2024-07-18
Ensure WordPress template hierarchy is intact for hybrid themes

ssang:
While working on a theme with the changes from #382 in place, I realized the issue was greater than just where the `index.php` file is placed in the hierarchy. Currently FSE templates are intentionally prioritized over blade templates, but we lose the WP Template hierarchy.

If we take a look at a Product CPT, this would be the template hierarchy that is currently in place.

image

As you can see, no matter how specific of a blade template file I create, it will never be prioritized over a `single.html` file in the templates directory. For hybrid themes, templates need to first be grouped by the file slugs with the FSE templates prioritized within those groupings.

This would be the result

image

### DIFF
--- a/src/Roots/Acorn/Sage/Concerns/FiltersTemplates.php
+++ b/src/Roots/Acorn/Sage/Concerns/FiltersTemplates.php
@@ -2,6 +2,8 @@
 
 namespace Roots\Acorn\Sage\Concerns;
 
+use Illuminate\Support\Str;
+
 trait FiltersTemplates
 {
     /**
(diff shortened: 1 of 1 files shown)


Log1x:
Thank you!

andronocean:
Is it possible to get this into a 4.x release while we wait for Acorn 5? This is a persistent hassle for hybrid themes — index.html taking precedence over home.blade.php puzzled me to no end a couple months ago, and just bit me again on another project.

PR #398, opened 2024-08-24
🧑‍💻 Append the original exception message to the skipped provider exception

Log1x:


montchr (commented):
Thank you!  I've been using this in an extended `Application` class and it is a vast improvement over the current behavior :slightly_smiling_face:. Eventually, it would be great to expose a stack trace from the problematic provider too.

QWp6t:
> Thank you! I've been using this in an extended `Application` class and it is a vast improvement over the current behavior 🙂. Eventually, it would be great to expose a stack trace from the problematic provider too.

You should be able to get a stack trace because it passes the the encountered exception as the `$previous` argument. Is that not working?

Log1x:
> You should be able to get a stack trace because it passes the the encountered exception as the `$previous` argument. Is that not working?

No, it doesn't work which is why I did this fix. It's otherwise a bit frustrating and has caused a good bit of support requests since the provider skipping behavior was implemented.

PR #402, opened 2024-09-24
Correctly split headers without leading whitespace

### LINKED ISSUE #401: ErrorException: `Undefined array key 1` with `header('x-test:header');`
### Version

4.3.0

I expected a response with status code 200.

### What actually happens?

I got an ErrorException: `Undefined array key 1`.

The exception happens on this line:
https://github.com/roots/acorn/blob/dce6d72c0f9eb1000cc0bc73e22cc2958cb08271/src/Roots/Acorn/Application/Concerns/Bootable.php#L146

It's triggered by the wp-seopress plugin because it adds the header `x-robots-tag:noindex, follow` (note the missing whitespace after the colon).
Here's the source line from the plugin: https://github.com/wp-seopress/wp-seopress-public/blob/d0d5c0cebc4cf238a96e7a3163fa7260aec43c32/src/Services/Sitemap/Headers.php#L23

MDN specifies
> **HTTP headers** let the client and the server pass additional information with an HTTP request or response. An HTTP header consists of its case-insensitive name followed by a colon (`:`), then by its value. [Whitespace](https://developer.mozilla.org/en-US/docs/Glossary/Whitespace) before the value is ignored.
See: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers

Here's the related RFC section: https://datatracker.ietf.org/doc/html/rfc7230#section-3.2
> Each header field consists of a case-insensitive field name followed by a colon (":"), optional leading whitespace, the field value, and optional trailing whitespace.

A previous RFC included more whitespace characters than the current RFC: https://datatracker.ietf.org/doc/html/rfc2616#section-4.2
So I think using PHPs ltrim function should be fine.

I suggest to update the Acorn code from
```diff
- [$header, $value] = explode(': ', $header, 2);
+ [$header, $value] = explode(':', $header, 2);
+ // HTTP/1.1 Header specification: https://datatracker.ietf.org/doc/html/rfc7230#section-3.2
+ // remove the optional leading whitespace
+ $value = ltrim($value);
```

I'm not sure if we even should use `trim` to also trim the optional trailing whitespace. I think we can ignore it.

### Steps to reproduce

Send a custom header without a space after the colon.

I can do this in my project by adding a `header-bug.php` with the following contents inside my `mu-plugins` folder:
```php
<?php

header('x-test:header');
```

### System info

_No response_

### Log output

_No response_

Yes

RafaelKr:
Fixes #401.

According to https://datatracker.ietf.org/doc/html/rfc7230#section-3.2 headers may contain optional leading whitespace before the value. The previous implementation threw an error if there was no whitespace.

### DIFF
--- a/src/Roots/Acorn/Application/Concerns/Bootable.php
+++ b/src/Roots/Acorn/Application/Concerns/Bootable.php
@@ -143,7 +143,10 @@ trait Bootable
     {
         Route::any('{any?}', fn () => tap(response(''), function (Response $response) {
             foreach (headers_list() as $header) {
-                [$header, $value] = explode(': ', $header, 2);
+                [$header, $value] = explode(':', $header, 2);
+                // HTTP/1.1 Header specification: https://datatracker.ietf.org/doc/html/rfc7230#section-3.2
+                // remove the optional leading whitespace
+                $value = ltrim($value);
 
                 if (! headers_sent()) {
                     header_remove($header);


src/Roots/Acorn/Application/Concerns/Bootable.php
```diff
@@ -143,7 +143,10 @@ protected function registerDefaultRoute(): void
     {
         Route::any('{any?}', fn () => tap(response(''), function (Response $response) {
             foreach (headers_list() as $header) {
-                [$header, $value] = explode(': ', $header, 2);
+                [$header, $value] = explode(':', $header, 2);
+                // HTTP/1.1 Header specification: https://datatracker.ietf.org/doc/html/rfc7230#section-3.2
+                // remove the optional leading whitespace
+                $value = ltrim($value);
```
QWp6t, line 149: I don't see the value in trimming the whitespace here.
```suggestion
```

QWp6t:
Thank you! 🙏

RafaelKr:
@QWp6t No problem!

I built a little Benchmark: https://github.com/RafaelKr/php-header-split-bench
Usage and the output/results I got are documented in the README.md.

As you can see by just using `explode` the header values are actually modified because one leading space is added.
Usually (and according to RFC) this shouldn't be a problem, but it definitely modifies the current behavior:
```json
{
    "X-With-Leading-Space": " Test: 123",
    "X-With-Leading-Space-And-Colon": " Test: 123",
    "X-Without-Leading-Space": "Test-Header",
    "Content-Type": " application\/x-www-form-urlencoded",
    "Cache-Control": " private, must-revalidate"
}
```
The same behavior can also be observed in real usage with Laravel/Acorn.

I think using `preg_split` wouldn't really impact performance but provide consistent output independent of leading whitespace.

Issue #404, opened 2024-10-02
PR #402 split headers follow-up discussion


RafaelKr:
### Summary

Last week I added a comment to #402. I added a comment to the now-merged PR where I'd like to discuss some more because I'm not sure if modifying current behaviour (as the PR currently does) is a good idea:

> I built a little Benchmark: https://github.com/RafaelKr/php-header-split-bench
> Usage and the output/results I got are documented in the README.md.
> 
> As you can see by just using `explode` the header values are actually modified because one leading space is added.
> Usually (and according to RFC) this shouldn't be a problem, but it definitely modifies the current behavior:
> ```json
> {
>     "X-With-Leading-Space": " Test",
>     "X-With-Leading-Space-And-Colon": " Test: 123",
>     "X-Without-Leading-Space": "Test-Header",
>     "Content-Type": " application\/x-www-form-urlencoded",
>     "Cache-Control": " private, must-revalidate"
> }
> ```
> The same behavior can also be observed in real usage with Laravel/Acorn.
> 
> I think using `preg_split` wouldn't really impact performance but provide consistent output independent of leading whitespace.

cc @QWp6t

QWp6t:
Hey thanks or the feedback!

These are all valid HTTP headers
```
x-test:test
x-test: test
x-test:  test
x-test:                        test
```
My original thought was that if a developer adds a header with leading whitespace, we should keep it intact. Removing leading whitespace would "definitely modify current behavior", as you put it.

Just noting that what you're proposing would override what might potentially be a developer's preference for deliberately including leading whitespaces in their headers. I don't see what value that adds on our end or for our users other than to serve an arbitrary aesthetic preference.

This isn't a strong opinion, though. I can definitely be convinced otherwise. Do you know if WordPress, Laravel, or Symfony already strip leading whitespace in header values?

RafaelKr:
Ah, I think we got a misunderstanding. The new problem now is that we're now always **adding** one space.

https://github.com/roots/acorn/blob/ef4a8a2f2719a063dc01c380a78633f3ea46b5db/src/Roots/Acorn/Application/Concerns/Bootable.php#L145-L153

Let's break this down with an example if someone did `header('Content-Type: image/png')` (notice this is what most people will do, there's exactly one space):
```php
foreach (headers_list() as $header) {
   // $header = 'Content-Type: image/png'
   [$header, $value] = explode(':', $header, 2); 
   // $header = 'Content-Type'
   // $value = ' image/png'

[...]

PR #405, opened 2024-10-03
🩹 Trim one optional leading space from headers

RafaelKr:
In my last PR #402 @QWp6t removed the `ltrim`. This leads to one added leading space to every header value. See https://github.com/roots/acorn/issues/404#issuecomment-2391024852

This PR can handle the case `X-Header:Value` which #402 already fixed without adding or removing leading whitespace in other cases.

Examples:
```
'X-Header:Value' => ['X-Header', 'Value']
'X-Header: Value' => ['X-Header', 'Value']
'X-Header:  Value' => ['X-Header', ' Value']
'X-Header:   Value' => ['X-Header', '  Value']
```


QWp6t:
I appreciate your persistence and thanks for clarifying the issue. 🙏

PR #412, opened 2024-11-06
🚑 Fix responses always returning `200` on WordPress routes (Fixes #411)

### LINKED ISSUE #411: All HTTP responses return with status code 200
### Version

^5.0.0

I expect the following dummy endpoint to return status code 400, but whenever I test it via postman it returns status code 200. This code is on a fresh Bedrock+Sage instance, with no other plugins activated.

The bug only seems to be present if we are using ^5.0.0 Acorn, while on on Acorn ^4.3.0 the response status code from the following endpoint is 400.
```
add_action('rest_api_init', function () {
    register_rest_route('custom/v1', '/status', [
        'methods' => 'GET',
        'callback' => 'custom_status_endpoint',
    ]);
});

function custom_status_endpoint()
{
    $response = [
        'status' => 'ok',
    ];

    $rest_response = new WP_REST_Response($response);
    $rest_response->set_status(400);

    return $rest_response;
}
```

### What actually happens?

The endpoint returns status code 200, although the response code is explicitly set to 400.  (Throwing a generic exception, also returns status code 200).

### Steps to reproduce

1. Use Acorn ^5.0.0
2. Create a WordPress REST API route.
3. Make it explicitly throw an Exception or set the response status code to 4xx.
4. Hit the endpoint via Postman.
5. Status code is still 200, even though it should be 4xx.

### System info

- macOS 15.
- Macbook M1 Pro
- Herd Pro 

### Log output

_No response_

Yes

Log1x:
This will need back-ported to v4 for those using the experimental flag.

### DIFF
--- a/src/Roots/Acorn/Application/Concerns/Bootable.php
+++ b/src/Roots/Acorn/Application/Concerns/Bootable.php
@@ -180,11 +180,12 @@ trait Bootable
         $path = Str::finish($request->getBaseUrl(), $request->getPathInfo());
 
         $except = collect([
-            rest_url(),
             admin_url(),
             wp_login_url(),
             wp_registration_url(),
         ])->map(fn ($url) => parse_url($url, PHP_URL_PATH))->unique()->filter();
+
+        $api = parse_url(rest_url(), PHP_URL_PATH);
 
         if (
             Str::startsWith($path, $except->all()) ||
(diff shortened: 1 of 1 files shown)


QWp6t (commented):
LGTM! :shipit:

PR #413, opened 2024-11-12
Add Hashing support

KIKOmanasijev:
This PR enables the Hash facade. 

Bcrpyt and argon are available as hashing drivers.


Log1x:
Thanks for the PR!

We should be able to just require in `illuminate/hashing` and load the provider. The `Illuminate/Foundation` folder is a mirror of https://github.com/laravel/framework/tree/11.x/src/Illuminate/Foundation and gets updated using a script.

Log1x:
I went ahead and required this in on #419

PR #417, opened 2024-12-05
🔧 Add the `FoundationServiceProvider` to the default providers (Fixes #409)

### LINKED ISSUE #409: Request validation and signature macros are missing
### Version

4.2.2

I expected the request validation and signature macros to be added.

### What actually happens?

They are not.

I think that we need just need to copy `FoundationServiceProvider::registerRequestValidation()` and `FoundationServiceProvider::registerRequestSignatureValidation()`. 

AFAICT the functions work as expected as long as the macros are added.

### Steps to reproduce

Try calling `$request->validate()`.

### System info

_No response_

### Log output

_No response_

Yes

Log1x:
This doesn't appear to have any side effects nowdays.

### DIFF
--- a/src/Roots/Acorn/DefaultProviders.php
+++ b/src/Roots/Acorn/DefaultProviders.php
@@ -23,6 +23,7 @@ class DefaultProviders extends DefaultProvidersBase
      * The additional framework providers.
      */
     protected array $additionalProviders = [
+        \Illuminate\Foundation\Providers\FoundationServiceProvider::class,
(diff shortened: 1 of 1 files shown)


sliver37:
When setting wp_debug to true, I get a skipping provider exception for ParallelTestingServiceProvider (as it's not found), from inside that FoundationServiceProvider. Using acorn with a standalone theme.

Log1x:
Are you using the v5 beta? `illuminate/testing` should be a dependency.

PR #448, opened 2025-03-16
Refactor routing method to include new parameters.

boxybird:
Updated `withRouting` to accept `$pages` and `$health` parameters, enhancing flexibility for routing configuration. This change ensures better alignment with the application's extended routing needs.


Log1x:
good catch, thanks!

boxybird:
You're welcome!

Issue #444, opened 2025-03-10
No json response from laravel web routes


bedirhandev:
### Version

5.0.0

### What did you expect to happen?

I expect JSON back from a laravel web request:

**result response:**
`{"category":{"id":1,"Name":"test"}}`

### What actually happens?

I get a string back like this: 
`"{\"category\":{\"Name\":\"gfgfd\",\"id\":121}}<!DOCTYPE html>\n etc... "`

### Steps to reproduce

web.php:
`Route::resource('category', CategoryController::class)->except(['index', 'show']);`

typescript file:
```
apiClient({
        url: '/category',
        method: 'POST',
        data: { Name: category.Name },
      })
    .catch((error) => console.error('Something went wrong: ', error))
    .then((response: any) => {
      const { category } = response['data'] as { category: Category }
      onSubmit(category)
    })
```

category controller file:
```
// Return the created resource with an HTTP 201 (Created) status
return response()->json(['category' => $category], 201);
```

**Solution:**:
```
if ($route->getName() !== 'wordpress') {
        add_action('parse_request', function() use ($request) { 
            $this->handleRequest($request);
            exit(0); // This had to be added used as example: (https://github.com/roots/acorn/blob/v5.0.0-beta.2/src/Roots/Acorn/Application/Concerns/Bootable.php#L242)
        });

        return;
    }
```

in the newer one: https://github.com/roots/acorn/blob/main/src/Illuminate/Foundation/Application.php#L1216 this is not there.

### Log output

```shell

```

### Please confirm this isn't a support request.

No

Log1x:
Hey! Thanks for the report. Care to do a PR with your solution?

bedirhandev:
Hey, it's now night here I will check it later and if possible do a PR. 

Thanks for the quick reply!

PR #450, opened 2025-03-16
Terminate request (wordpress route) after request is handled.

bedirhandev:
Fix of issue: https://github.com/roots/acorn/issues/444


Log1x:
thanks!

PR #464, opened 2025-05-09
Fix app registration callbacks not being called (Fixes #463)

### LINKED ISSUE #463: App registration callbacks are not called as expected
### Version

5.0.3

I'd expect callbacks added via `Application::configure()->registered()` to be called after all providers have been registered.

### What actually happens?

They are never called.

The issue is caused by [this line](https://github.com/roots/acorn/blob/e8c787db932fd959f53809e153733e73bc4e42dc/src/Illuminate/Foundation/Application.php#L873) not being present in [this override](https://github.com/roots/acorn/blob/e8c787db932fd959f53809e153733e73bc4e42dc/src/Roots/Acorn/Application.php#L322). It looks like this functionality was just missed when the upgrade to Laravel v11 was made.

`booting()` and `booted()` work as intended.

### Steps to reproduce

1. Add `->registered(function () { Log::info('App registration is now done.'); })` to your `Application::configure()` call.
2. Boot your app
3. Find that the message `App registration is now done.` has not been logged.

### System info

Ubuntu 24.04

### Log output

```shell

```

No

stefanfisk:


Log1x:
nice! thank you

Issue #406, opened 2024-10-07
Experimental Request Handler breaks WP Rocket


RafaelKr:
### Version

4.3.0

### What did you expect to happen?

In a Radicle project we have the experimental Request handler enabled via `ACORN_ENABLE_EXPERIMENTAL_WORDPRESS_REQUEST_HANDLER=true`.

We now installed WP Rocket (Previously we had WP Optimize where it also didn't work).

It should create static html files inside `public/content/cache/wp-rocket/example.com`

### What actually happens?

It only creates those static html files when the experimental request handler is disabled.

It's related to output buffering. WP Rocket receives an empty output buffer at https://github.com/wp-media/wp-rocket/blob/09708aa55d362a9535ad4d45cb68280325efaecb/inc/classes/Buffer/class-cache.php#L259

The order of calls is:
- public/content/advanced-cache.php (generated) calls [wp-media/wp-rocket@09708aa/inc/classes/Buffer/class-cache.php::maybe_init_process()](https://github.com/wp-media/wp-rocket/blob/develop/inc/classes/Buffer/class-cache.php#L81)
- `maybe_init_process` calls [`ob_start( [ $this, 'maybe_process_buffer' ] );`](https://github.com/wp-media/wp-rocket/blob/09708aa55d362a9535ad4d45cb68280325efaecb/inc/classes/Buffer/class-cache.php#L157)
- Some time later `maybe_process_buffer` is called with an empty string https://github.com/wp-media/wp-rocket/blob/09708aa55d362a9535ad4d45cb68280325efaecb/inc/classes/Buffer/class-cache.php#L256

When we disable the experimental request handler the buffer contains the correct content and the static file is generated.

### Steps to reproduce

- Add `ACORN_ENABLE_EXPERIMENTAL_WORDPRESS_REQUEST_HANDLER=true` to `.env` 
- `composer install wp-media/wp-rocket`
- Activate WP Rocket (By default it will add `define( 'WP_CACHE', true ); // Added by WP Rocket` to wp-config.php, but that should be fine for testing. Can be disabled via https://docs.wp-rocket.me/article/958-how-to-use-wp-rocket-with-bedrock-wordpress-boilerplate)
- Go to `public/content/cache/wp-rocket/your-domain.localhost` (or the wp-content path which is configured without Radicle)
- Load a page and see that there are no generated files
- Remove `ACORN_ENABLE_EXPERIMENTAL_WORDPRESS_REQUEST_HANDLER=true` from `.env` 
[...]

PR #480, opened 2025-08-25
🩹 fix: Clean output buffering up to initial level

tombroucke:
This PR should resolve https://github.com/roots/acorn/issues/406

The `registerWordPressRoute` method deletes all levels of output buffering. Caching plugins try to buffer the whole output and write that to a file. Output buffering is started by caching plugins in `{wp-content|app}/advanced-cache.php`, and saved in a callback for that `ob_start()`. As acorn is deleting all buffers from all levels in `Bootable::registerWordPressRoute()`, the output is empty by the time the caching plugin's callback is called.

With this PR, we get the initial ob level,  and delete buffers up until this level.

I have tested this fix with
- WP Rocket
- [Super Page Cache](https://wordpress.org/plugins/wp-cloudflare-page-cache/)
- https://roots.io/acorn/docs/routing/#basic-routing-example
- wordpress routing + livewire
and didn't encounter any issues, but please test thoroughly, as this is quite a critical change.

### DIFF
--- a/src/Roots/Acorn/Application/Concerns/Bootable.php
+++ b/src/Roots/Acorn/Application/Concerns/Bootable.php
@@ -123,7 +123,7 @@ trait Bootable
         $kernel->bootstrap($request);
 
         if ($this->app->handlesWordPressRequests()) {
-            $this->registerWordPressRoute();
+            $this->registerWordPressRoute(ob_get_level());
         }
 
         try {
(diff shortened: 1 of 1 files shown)


tombroucke:
This also fixes #423

QWp6t (approved):
I like it! Thanks! 🙏

Issue #481, opened 2025-09-02
JSON option in about command incompatible with WP-CLI


dsturm:
### Version

5.0.5

### What did you expect to happen?

When running `wp acorn about --json` we should get the application information as a json-formatted string. This is likely to be used with [Laravel Herd](https://herd.laravel.com/docs/macos/extending-herd/custom-drivers) to show the data in the app.

### What actually happens?

The command exits with `The "--format" option does not exist.`.

This is because WP-CLI converts the `--json` to `--format=json` and unsets it afterwards in [php/WP_CLI/Runner.php](https://github.com/wp-cli/wp-cli/blob/ae3545998aa8d2197be3c040bcbee82c4e0e7e66/php/WP_CLI/Runner.php#L896) before the orginal AboutCommand class is executed.

### Steps to reproduce

1. Run `wp acorn about --json` in CLI
2. The code exits with `The "--format" option does not exist.`

### Log output

```shell
[object] (Symfony\Component\Console\Exception\RuntimeException(code: 0): The "--format" option does not exist. at /Users/dcsturm/wrk/sites/domain/web/app/themes/sage/vendor/symfony/console/Input/ArgvInput.php:226)
[stacktrace]
#0 /Users/dcsturm/wrk/sites/domain/web/app/themes/sage/vendor/symfony/console/Input/ArgvInput.php(153): Symfony\Component\Console\Input\ArgvInput->addLongOption('format', 'json')
#1 /Users/dcsturm/wrk/sites/domain/web/app/themes/sage/vendor/symfony/console/Input/ArgvInput.php(88): Symfony\Component\Console\Input\ArgvInput->parseLongOption('--format=json')
#2 /Users/dcsturm/wrk/sites/domain/web/app/themes/sage/vendor/symfony/console/Input/ArgvInput.php(77): Symfony\Component\Console\Input\ArgvInput->parseToken('--format=json', true)
#3 /Users/dcsturm/wrk/sites/domain/web/app/themes/sage/vendor/symfony/console/Input/Input.php(53): Symfony\Component\Console\Input\ArgvInput->parse()
#4 /Users/dcsturm/wrk/sites/domain/web/app/themes/sage/vendor/symfony/console/Command/Command.php(276): Symfony\Component\Console\Input\Input->bind(Object(Symfony\Component\Console\Input\InputDefinition))
[...]

PR #482, opened 2025-09-02
Add --format option to fix --json option with WP-CLI

dsturm:
To fix the current issue, where WP-CLI overrides the `--json` option before executing the about command (#481), I came up with adding a third option `--format` to the command, which conditionally sets the `json` option to true.


QWp6t (approved):
this is a good workaround until one day when https://github.com/wp-cli/wp-cli/pull/6113 maybe gets merged lol

PR #483, opened 2025-09-24
Quickfix Acorn\PackageManifest: ${composerPath} -> {$composerPath}

sybbear:
Hello,
this makes 3 small fixes that stops php from producing 
`[WARNING: Using ${var} in strings is deprecated, use {$var} instead]`

Please make a new release, if it is easy to do on this 2.x branch, so we can enjoy cleaner logs

The same issue is present with version `3`, however since version `4` seems fixed (search by `${`)

Best regards

### DIFF
--- a/src/Roots/Acorn/PackageManifest.php
+++ b/src/Roots/Acorn/PackageManifest.php
@@ -64,7 +64,7 @@ class PackageManifest extends FoundationPackageManifest
         $packages = array_reduce($this->composerPaths, function ($all, $composerPath) {
             $packages = [];
 
-            $path = "${composerPath}/vendor/composer/installed.json";
+            $path = "{$composerPath}/vendor/composer/installed.json";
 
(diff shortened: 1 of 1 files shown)


sybbear:
Any chance to merge this?
@retlehs or someone for the team

Totally harmless fix, however significantly reduces log noise

Log1x:
Is there any particular reason you can't upgrade? The upgrade path should be pretty smooth.

sybbear:
@retlehs @Log1x 
Roots/acorn 3 requires L9 and version 4 requires L10
We have a pretty large system that relies on L8 and we can't upgrade it of now due to vendor constraints.

Is there something else we could do other than forking this? 
I'd write tests for this PR, but there are no tests to write for these kind of changes, since the logic is exactly the same, just correct string interpolation

If you could possibly consider 🙏 😄

sybbear:
Big thank you!

PR #485, opened 2025-10-16
🩹 fix: Relay status code set by WordPress

### LINKED ISSUE #484: `->withRouting(wordpress: true)` serves 200 status code on non-existent pages with trailing slash
### Version

v5.0.5

Server 404 should be the correct status code on `/non-existent/` URLs

### What actually happens?

Server 200 status on `/non-existent/` URLs

Note: The correct 404 status code returns without the trailing slash: `/non-existent`

### Steps to reproduce

1. Add `->withRouting(wordpress: true)` to the bootloader:

```php
Application::configure()
    ->withProviders([
        App\Providers\ThemeServiceProvider::class,
    ])
    ->withRouting(wordpress: true)
    ->boot();
```

2. Check the headers of a non-existent page with a trailing slash at the end

### System info

_No response_

### Log output

```shell

```

Yes

tombroucke:
Fixes #484 
WordPress sets a 404 statuscode in the `WP::handle_404()` method, but that got overwritten by the default status code of the Response class. We can fetch the status set by WP using `http_response_code()`, and pass that in to the `sendHeaders()` method. This will relay all status codes, but I'm assuming if WP throws a 301, we should follow that.

Tests are running fine, but this should probably be tested manually.

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,4 +1,4 @@
-ARG PHP_IMAGE=8-fpm-bookworm
+ARG PHP_IMAGE=8.5-fpm-bookworm
 FROM php:${PHP_IMAGE}
 
 LABEL maintainer="Roots Team"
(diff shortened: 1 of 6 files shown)


retlehs (approved):
```
❯ curl -I http://example.test/asdf/
HTTP/1.1 404 Not Found
```

Thank you!

Log1x:
nice find! does https://github.com/roots/acorn/blob/8f4d458f9d87c6f2cfb433468370d0d32b7eb0d2/src/Roots/Acorn/Application/Concerns/Bootable.php#L166 still end up being necessary?

PR #489, opened 2025-11-28
Fix PHP 8.5+ deprecation for PDO::MYSQL_ATTR_SSL_CA

whygareth:
### Summary
This PR addresses the deprecation warning for `PDO::MYSQL_ATTR_SSL_CA` introduced in PHP 8.4 (and present in PHP 8.5 builds).

### The Issue
In newer PHP versions, the `PDO::MYSQL_ATTR_SSL_CA` constant is deprecated in favor of the namespaced `Pdo\Mysql::ATTR_SSL_CA`. Using the old constant results in a deprecation notice.

### The Solution
I have updated `config/database.php` to use a conditional check. The configuration now uses the new `Pdo\Mysql::ATTR_SSL_CA` constant if it is defined, while falling back to the legacy `PDO::MYSQL_ATTR_SSL_CA` constant for older PHP versions to ensure backward compatibility.

### Affects
- MySQL connection
- MariaDB connection
- WordPress connection

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,4 +1,4 @@
-ARG PHP_IMAGE=8-fpm-bookworm
+ARG PHP_IMAGE=8.5-fpm-bookworm
 FROM php:${PHP_IMAGE}
 
 LABEL maintainer="Roots Team"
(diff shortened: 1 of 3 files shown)


Log1x:
Hey thanks for this! It looks like this was actually addressed in Laravel a few days ago. Would you want to copy over the exact fix they did in https://github.com/laravel/laravel/blob/12.x/config/database.php ? Namely checking the PHP version.

whygareth:
Good call! Thanks for the heads up.
I've updated the check to use PHP_VERSION_ID, but I stuck with 80400 (PHP 8.4) instead of 80500. Since the deprecation starts in 8.4, using 80500 would sadly leave those users with the warning still active.

QWp6t (approved):

PR #500, opened 2026-02-25
🩹 Skip env file loading when no .env file exists

### LINKED ISSUE #496: error reading env file, if no env file present
### Version

5.0.2

When no `.env` file exists anywhere in the directory tree , Acorn should handle this gracefully without triggering a PHP warning.

We provide our environment variables to the server using our GKE Setup without a `.env`

### What actually happens?

A PHP warning is emitted on every request:

> Warning: file_get_contents(/path/to/web/app/themes/my-theme/.env): Failed to open stream: No such file or directory
>   in /path/to/web/app/themes/my-theme/vendor/vlucas/phpdotenv/src/Store/File/Reader.php on line 73

The warning is suppressed by phpdotenv's `@` operator, but our Sentry Instance still captures it, generating noise in error tracking.

### Steps to reproduce

1. Set up a WordPress project where environment variables are provided by the server (e.g. via ddev's .ddev/.env, Docker env vars, or server-level config) — no .env file in the project tree
2. Install Acorn 5.0.2 in the theme
3. Load any page
4. Observe the warning in your error tracking tool or by temporarily removing the @ from phpdotenv's Reader.php:73

 Root cause:

  In Application.php:107-112, environmentPath() uses Filesystem::closest() to walk up from basePath() looking for a .env file. When none is found, it falls back to $this->basePath() (the theme directory):

  public function environmentPath(): string
  {
      return is_file($envPath = (new Filesystem)->closest($this->basePath(), '.env') ?? '')
          ? dirname($envPath)
          : $this->basePath();
  }

  Later, phpdotenv's Reader::readFromFile() calls @file_get_contents() on the non-existent <theme-dir>/.env, producing the suppressed warning.

  Workaround:

As for now we consider to create an empty .env file so that the error is not happening.

### System info

Happens locally on a ddev environment and on a GKE production environment, so likely irrelevant.

### Log output

```shell

```

Yes

apermo:
## Summary

Fixes #496

When no `.env` file exists anywhere in the directory tree, `Application::environmentPath()` falls back to `basePath()`. phpdotenv's `Reader::readFromFile()` then calls `@file_get_contents()` on the non-existent `<basePath>/.env`, emitting a suppressed PHP warning. Error tracking tools like Sentry still capture these suppressed warnings, generating noise on every request.

### Approach

This adds a `Roots\Acorn\Bootstrap\LoadEnvironmentVariables` bootstrapper — following the existing pattern used by `LoadConfiguration`, `HandleExceptions`, and `RegisterFacades` — that checks if the environment file exists before attempting to load it. If the file doesn't exist, loading is skipped entirely since environment variables are provided by other means (server config, Docker, GKE, etc.).

Both `Http\Kernel` and `Console\Kernel` are updated to use the new Acorn bootstrapper.

### Changes

- **New**: `src/Roots/Acorn/Bootstrap/LoadEnvironmentVariables.php` — extends `Illuminate\Foundation\Bootstrap\LoadEnvironmentVariables` with an `is_file()` guard
- **Updated**: `src/Roots/Acorn/Http/Kernel.php` — use `Roots\Acorn\Bootstrap\LoadEnvironmentVariables`
- **Updated**: `src/Roots/Acorn/Console/Kernel.php` — use `Roots\Acorn\Bootstrap\LoadEnvironmentVariables`

No vendored `src/Illuminate/` files are modified.

### DIFF
--- a/tests/Bootstrap/LoadEnvironmentVariablesTest.php
+++ b/tests/Bootstrap/LoadEnvironmentVariablesTest.php
@@ -1,0 +1,60 @@
+<?php
+
+use Illuminate\Contracts\Foundation\Application;
+use Roots\Acorn\Bootstrap\LoadEnvironmentVariables;
+use Roots\Acorn\Tests\Test\TestCase;
+
+uses(TestCase::class);
+
+class TestableLoadEnvironmentVariables extends LoadEnvironmentVariables
+{
+    public bool $dotenvLoaded = false;
+
+    protected function createDotenv($app)
+    {
+        $this->dotenvLoaded = true;
+
+        $dotenv = Mockery::mock(\Dotenv\Dotenv::class);
+        $dotenv->shouldReceive('safeLoad')->once();
+
+        return $dotenv;
+    }
+}
+
+function createEnvAppMock(string $envPath, string $envFile = '.env', bool $cached = false): Application
+{
+    $app = Mockery::mock(Application::class);
+    $app->shouldReceive('configurationIsCached')->andReturn($cached);
+    $app->shouldReceive('environmentPath')->andReturn($envPath);
+    $app->shouldReceive('environmentFile')->andReturn($envFile);
+    $app->shouldReceive('runningInConsole')->andReturn(false);
+
+    return $app;
+}
+
+it('should skip loading when env file does not exist', function () {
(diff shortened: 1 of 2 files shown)


apermo:
Thank you for the quick reply, I'll look for another option. And come back with an updated version.

PR #507, opened 2026-03-08
🐛 Force root URL to home_url() for correct route URL generation

### LINKED ISSUE #400: route() returns incorrect URLs when request is not handled by index.php
### Version

v4.3.0

I expected `route()` to return correct URLs whenever used.

### What actually happens?

When used outside of the frontend or console `route()` prefixes the URLs with whatever `.php` file WordPress handled the request via. 

So `http://localhost/my-route` becomes `http://localhost/wp/wp-admin/admin-post.php/my-route` etc. 

The issue can be solved by calling `Illuminate\Support\Facades\URL::forceRootUrl(home_url())` during boot but I am not sure about possible caveats.

### Steps to reproduce

1. Add `Route::get('/test')->name('test');` to `routes/web.php`
2. Add `add_action('admin_post_nopriv_route_test', function() { echo route('test'); });` to `functions.php`
3. Open `/wp-admin/admin-post.php?action=route_test` as appropriate for your local environment

### System info

 Pop!_OS 22.04 LTS

### Log output

_No response_

Yes

retlehs:
## Summary
- Forces `URL::forceRootUrl(home_url())` during HTTP boot so `route()` generates correct URLs regardless of the PHP entry point
- Without this, requests handled by `admin-post.php`, `admin-ajax.php`, `wp-cron.php`, etc. cause Symfony's `Request::prepareBaseUrl()` to derive the base path from `SCRIPT_NAME`, resulting in URLs like `http://example.com/wp/wp-admin/admin-post.php/my-route`
- Reproduced and verified the fix with `admin-post.php`, `admin-ajax.php`, and `wp-cron.php`

Closes #400

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/Roots/Acorn/Application/Concerns/Bootable.php
+++ b/src/Roots/Acorn/Application/Concerns/Bootable.php
@@ -122,6 +122,8 @@ trait Bootable
 
         $kernel->bootstrap($request);
 
+        \Illuminate\Support\Facades\URL::forceRootUrl(home_url());
+
         if ($this->app->handlesWordPressRequests()) {
             $this->registerWordPressRoute(ob_get_level());
         }


redalpha01:
This one seems to cause issues when using wpml and wanting to show routes from multiple locales

If I'm in /fr and want to show a /en link, I don't want it to be /fr/en

retlehs:
Please open a new issue with reproduction steps

PR #514, opened 2026-03-09
🐛 Isolate cache, session, and queue data on multisite

### LINKED ISSUE #501: Multisite: shared cache/storage directory causes cross-site data contamination
## Description

Acorn's storage/cache directory (`web/app/cache/acorn`) is shared across all sites in a WordPress multisite installation. This means that `wp acorn optimize` and normal runtime caching overwrite each other, causing the last site that was optimized or visited to dictate the cached state for all other sites.

## Problem

In a multisite environment, all subsites write to the same storage path. This leads to several issues:

- **View cache collisions** — `wp acorn optimize` compiles views into a single shared directory. If a plugin registers a view path on only a subset of sites, the compiled cache from one subsite can be served to another where that plugin isn't active.
- **Cache data leaking between sites** — Sites sharing the same cache keys (via Laravel's cache system) can receive data intended for a different subsite, depending on which site was visited first.
- **Session cross-contamination** — If using the filesystem session driver, sessions are stored in a shared directory across all subsites.

These issues likely extend to any storage-backed feature that assumes a single-site context.

## Expected Behavior

Each subsite in a multisite installation should have isolated storage paths (cache, views, sessions, logs) so that one site's cached data never leaks into another.

## Workaround

The following snippet can be placed in the theme's `functions.php` before the application boots to isolate storage per site:

```php
if (is_multisite()) {
    $storage = Str::finish(WP_CONTENT_DIR, '/cache/acorn/site-' . get_current_blog_id());

    foreach (['framework/cache/data', 'framework/views', 'framework/sessions', 'logs'] as $dir) {
        (new Filesystem)->ensureDirectoryExists("{$storage}/{$dir}");
    }

    define('ACORN_STORAGE_PATH', $storage);
}
```

This creates per-site storage directories like `web/app/cache/acorn/site-1`, `web/app/cache/acorn/site-2`, etc.

## Additional Considerations

- **Non-filesystem cache drivers** — The workaround above only addresses the filesystem driver. For cache drivers that support tags (e.g. Redis, Memcached), per-site cache tags would be a more performant approach and allow invalidating all cache for a single site at once. For drivers without tag support, cache key prefixing per site would be the fallback.
- **Non-filesystem session drivers** — The same cross-contamination can occur with non-filesystem session drivers and would need similar per-site isolation.
- **View compilation** — There may be architectural changes possible that avoid per-site cache paths for compiled views entirely.

## Environment

- WordPress multisite
- Acorn (latest)
- Observed via `wp acorn optimize` and runtime cache behavior

## References

 - https://discord.com/channels/461985401494700032/819441934761197598/1478078676443594854

----

💫 This issue was generated by robots.

retlehs:
## Summary
- On multisite networks, all sites shared the same cache keys, session cookies, and queue job context, causing data collisions and cross-contamination
- Dynamically prefixes `cache.prefix` with `blog_{id}_` and appends `_{id}` to `session.cookie` at runtime in `AcornServiceProvider`
- Injects `blogId` into every queued job payload and switches to the correct blog context before the job is processed, restoring it afterward
- Hooks into `switch_blog` action so config stays correct when blog context changes (e.g. queue workers processing jobs for multiple sites)
- Works with all cache/session drivers (file, Redis, Memcached, database, etc.) and all queue drivers
- Views, logs, and storage paths remain shared as expected for multitenant Laravel applications

## Test plan
- [x] Unit tests for cache prefix and session cookie isolation per blog
- [x] Unit test confirming single-site config is unchanged
- [x] Unit test for config update on blog switch
- [x] Unit test for blogId injection into queue payloads
- [x] Unit test for blog context switching on job processing
- [x] Unit test for blog context restore after job exception
- [x] Unit test for graceful handling of jobs without blogId
- [x] Verified on a local Bedrock multisite (2 subsites) — each site gets unique cache prefix and session cookie
- [x] End-to-end queue test with database driver — dispatched jobs from both sites, processed with `queue:work`, confirmed correct blog context via log output
- [x] All 105 tests passing

Closes #501

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/Roots/Acorn/Providers/AcornServiceProvider.php
+++ b/src/Roots/Acorn/Providers/AcornServiceProvider.php
@@ -186,22 +186,28 @@ class AcornServiceProvider extends ServiceProvider
         });
 
         $events = $this->app->make('events');
+        $switchedJobs = [];
 
-        $events->listen(JobProcessing::class, function (JobProcessing $event) {
+        $events->listen(JobProcessing::class, function (JobProcessing $event) use (&$switchedJobs) {
             $payload = $event->job->payload();
 
             if (isset($payload['blogId'])) {
                 switch_to_blog((int) $payload['blogId']);
+                $switchedJobs[spl_object_id($event->job)] = true;
             }
         });
 
-        $events->listen(JobProcessed::class, function () {
-            restore_current_blog();
-        });
+        $restore = function ($event) use (&$switchedJobs) {
+            $id = spl_object_id($event->job);
 
-        $events->listen(JobExceptionOccurred::class, function () {
-            restore_current_blog();
-        });
+            if (isset($switchedJobs[$id])) {
+                restore_current_blog();
+                unset($switchedJobs[$id]);
+            }
+        };
+
+        $events->listen(JobProcessed::class, $restore);
+        $events->listen(JobExceptionOccurred::class, $restore);
     }
 
     /**
--- a/tests/Providers/MultisiteIsolationTest.php
+++ b/tests/Providers/MultisiteIsolationTest.php
@@ -16,6 +16,7 @@ beforeEach(function () {
     $this->stub('is_multisite', fn () => $this->isMultisite);
     $this->stub('get_current_blog_id', fn () => $this->blogId);
     $this->stub('switch_to_blog', function ($id) {
+        $this->blogIdStack[] = $this->blogId;
         $this->blogId = $id;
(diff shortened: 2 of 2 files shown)


QWp6t (approved):
This theoretically looks good, but I haven't test with an actual multisite setup. I'm happy enough with it that I think we can safely merge it.

PR #543, opened 2026-05-27
🙈 Ignore IDE files

rafaucau:
This makes contributing less annoying by ignoring local IDE settings.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,5 @@
+.idea
+.vscode
 .php_cs.cache
 .phpunit.*
 phpunit.xml


QWp6t:
I'm not strictly opposed to this, but we've gone several years at this point without this so I'm not fully sold on the need for it either.

That said, you'll encounter similar situations when contributing to other repos out there. Not everyone is going to have a .gitignore rule that suits your specific tools, and you can self-serve with a global `.gitignore` or a local `.git/info/exclude`.

I'll wait for others to weigh in to see what they think.

retlehs:
Thanks @rafaucau! We'd rather keep editor-specific entries out of the repo

PR #544, opened 2026-05-29
fix: use home_url() for multisite-aware asset paths in @vite directive

ivowk:
In a WordPress multisite network, the @vite directive generates asset tags pointing to the main site's domain instead of the current subsite's domain, causing CORS errors on subsites. This issue has also been discussed on [Roots Discourse](https://discourse.roots.io/t/incorrect-domain-when-loading-js-assets-in-multisite/30113).

```
Access to script at 'https://main.example.com/app/themes/my-theme/public/build/assets/app.js'
from origin 'https://sub.example.com' has been blocked by CORS policy.
```

Vite::assetPath() resolves the URI from assets.manifests.*.url, which is set once at boot from get_theme_file_uri() and never reflects the active subsite context.

`home_url()` returns the correct origin for the currently active site. The fix is scoped to `is_multisite()` so single-site installs  are completely unaffected.


retlehs:
Thanks for the fix. Could you update this to avoid `home_url($path)`? It breaks subdirectory multisite by inserting the subsite slug into the asset path

Ideally rewrite only the stale host, preserve the path/query, and add tests for subdomain + subdirectory multisite

ivowk:
@retlehs Updated, thank you for the feedback!

[assetPath()](cci:1://file:///Users/ivowermelinger/repos/acorn/src/Roots/Acorn/Assets/Vite.php:10:4-42:5) now uses a private [rewriteHost()](cci:1://file:///Users/ivowermelinger/repos/acorn/src/Roots/Acorn/Assets/Vite.php:28:4-54:5) helper that takes only the scheme/host/port from `home_url()` and applies them to the asset URI, while preserving the original path, query, and fragment. This fixes the subdirectory multisite issue where `home_url($path)` would prepend the subsite slug into the asset path.

See the changes here:
- [[src/Roots/Acorn/Assets/Vite.php](cci:7://file:///Users/ivowermelinger/repos/acorn/src/Roots/Acorn/Assets/Vite.php:0:0-0:0)](https://github.com/ivowk/acorn/blob/fix/vite-multisite-asset-url/src/Roots/Acorn/Assets/Vite.php)
- [[tests/Assets/ViteTest.php](cci:7://file:///Users/ivowermelinger/repos/acorn/tests/Assets/ViteTest.php:0:0-0:0)](https://github.com/ivowk/acorn/blob/fix/vite-multisite-asset-url/tests/Assets/ViteTest.php)

I also added tests for single-site passthrough, subdomain multisite, and subdirectory multisite.

src/Roots/Acorn/Assets/Vite.php
```diff
@@ -17,6 +17,43 @@ class Vite extends FoundationVite
...
+     * Swap the scheme/host/port of a URI to match home_url(), preserving the
+     * original path/query/fragment.
+     *
+     * Assumes assets are served from the same origin as the site.
+     *
+     * @param  string  $uri  The URI to rewrite.
+     * @param  string  $base  The base URL to use for the rewrite.
+     * @return string  The rewritten URI.
+     */
+    private function rewriteHost(string $uri, string $base): string
```
QWp6t:
```suggestion
    protected function rewriteHost(string $uri, string $base): string
```

